PR #7, opened 2015-10-24
Promise based API

rhysd:
This commit has big changes to API.

```js
var cp = require('child_process');
var attach = require('promised-neovim-client').attach;

var nvim_proc = cp.spawn('nvim', ['-u', 'NONE', '-N', '--embed'], {});
attach(nvim_proc.stdin, nvim_proc.stdout).then(function(nvim) {
  nvim.on('request', function(method, args, resp) {
    // handle msgpack-rpc request
  });

  nvim.on('notification', function(method, args) {
    // handle msgpack-rpc notification
  });

  nvim
    .command('vsp')
    .then(() => nvim.getWindows())
    .then(windows => {
      console.log(windows.length);  // 2
      console.log(windows[0] instanceof nvim.Window); // true
      console.log(windows[1] instanceof nvim.Window); // true
      return nvim.setCurrentWindow(windows[1])
        .then(() => nvim.getCurrentWindow())
        .then(win => {
          console.log(win.equals(windows[1]))  // true
          nvim.quit();
          nvim.on('disconnect', () => console.log("Nvim exited!"));
        });
    }).catch(err => console.log(err.message));

}).catch(err => console.log(err.message));
```

I changed the API to use `Promise` instead of callbacks.  I added new dependency `es6-promise` for ES6 Promise shim.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,5 +4,6 @@ before_install:
   - eval "$(curl -Ss https://raw.githubusercontent.com/neovim/bot-ci/master/scripts/travis-setup.sh) nightly-x64"
 
 node_js:
-- '0.12'
-- '0.10'
+  - 'stable'
+  - '0.12'
+  - '0.10'
--- a/README.mkd
+++ b/README.mkd
@@ -13,7 +13,7 @@ npm install --global neovim-client
 ## Usage
 
 This package exports a single `attach()` function which takes a pair of
-write/read streams and invokes a callback with a Nvim API object. This is
+write/read streams and returns a promise of Nvim API object. This is
 similar to [node-msgpack5rpc](https://github.com/tarruda/node-msgpack5rpc), but
 it provides an automatically generated API.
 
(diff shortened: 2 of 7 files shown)


tarruda:
There's no need to change the internal implementation to use promises, you can use a library like [bluebird](https://github.com/petkaantonov/bluebird) to wrap the existing API into a promise API using the internal metadata(which is already used to generate the typescript declarations).

Perhaps you'd like to try that approach and create a separate package that wraps this one into promises?

tarruda:
If you don't want to create another package, you can also modify this PR to add a new function that wraps the existing callback-based API into a promise-based API, but using another package(eg: neovim-client-promise) seems cleaner since it doesn't add dependency to a promise library for those that prefer a simple callback API. I was thinking of doing this eventually, but if you create another package I can create a repo in the org and add you as maintainer.

FWIW it is also possible to expose a synchronous API with [node-fibers](https://github.com/laverdet/node-fibers) in request/notification handlers, but I also think it belongs in a separate package. IMO we should take advantage of node.js excellent module/package system to simplify packages as much as possible.

rhysd:
Alright, yeah I already created [separate package](https://www.npmjs.com/package/promised-neovim-client) for my own use.  I agree your opinion and will close this issue.
Thank you for your review.

PR #14, opened 2016-06-15
Correctly define ui_* methods from neovim api info.

megalord:
Fixes https://github.com/neovim/node-host/issues/13

Defines  `attach`, `detach`, and `tryResize` on the `Nvim` type.  This is consistent with the [python client](https://github.com/neovim/python-client/blob/master/neovim/api/nvim.py#L172).  Also, the auto-generated methods are aliased to their old names to keep backwards compatibility.

I'm just getting started in neovim plugins, so please use a critical eye in reviewing the PR; the tests don't seem to cover a change like this.  I did try copying these changes into my install of the `node-host`, and it did fix the error with `UpdateRemotePlugins`.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -3,6 +3,9 @@ declare module "neovim-client" {
   function attach(writer: NodeJS.WritableStream, reader: NodeJS.ReadableStream, cb: (err: Error, nvim: Nvim) => void);
 
   interface Nvim {
+    attach(width: number, height: boolean, enable_rgb: (err: Error) => void, cb: (err: Error) => void): void;
+    detach(cb: (err: Error) => void): void;
+    tryResize(width: number, height: (err: Error, res: Object) => void, cb: (err: Error, res: Object) => void): void;
     command(str: string, cb: (err: Error) => void): void;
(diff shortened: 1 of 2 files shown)


megalord:
Ping @tarruda or @nhynes

nhynes:
0x06
Thanks for the bug report! I'll take a look at this a bit later today.

nhynes:
Is there any reason why we want to alias `uiAttach` to `attach`? This would be inconsistent with the Python client and could be confused with `attach` defined as the library entry point.

Perhaps, instead, we do
```diff
@@ -33,7 +33,7 @@ function generateWrappers(Nvim, types, metadata) {
-    var methodName = _.camelCase(parts.slice(1).join(('_')));
+    var methodName = _.camelCase(parts.slice(typeName !== 'Ui').join(('_')));
```
and entirely remove the corresponding assignments in `addExtraNvimMethods`. What do you think?

megalord:
Sorry I was offline all day yesterday, but thanks!  I actually like your suggestion a lot, but it would also complicate the logic around determining the `Type`.  It seems like the api info is a bit inconsistent here, because I would expect `ui_attach` to result in an `attach` method on a `Ui` object or the info to change to `vim_ui_attach`.

PR #18, opened 2016-08-09
Fix parameter types of ui*() functions and add missing return type

rhysd:
I found some mistakes in `index.d.ts` and fixed them.


justinmk:
LGTM, would like @nhynes or @tarruda to take a peak though.

tarruda:
LGTM

PR #19, opened 2016-08-10
Export attach() also as a default

rhysd:
Previously, `attach()` was exported as _whole_ of the package.  However, `index.d.ts` says that it is exported as _default_ of package.  _whole_ and _default_ is not the same thing in ES2015 and TypeScript, they are always confusing:

```javascript
// Whole import
import * as attach from 'neovim-client';

// Default import
import attach from 'neovim-client';
```

In general, people prefers default export because it's much more generic.  So I exported `attach()` also as a 'default' of this package.

### DIFF
--- a/generate-typescript-interfaces.js
+++ b/generate-typescript-interfaces.js
@@ -58,8 +58,7 @@ attach(proc.stdin, proc.stdout, function(err, nvim) {
 
   // use a similar reference path to other definitely typed declarations
   process.stdout.write('declare module "neovim-client" {\n');
-  process.stdout.write('  export default attach;\n');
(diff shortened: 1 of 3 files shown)


rhysd:
Without this PR, TypeScript definition file does not work because nothing is exported as default for now.  Can anyone review this?

nhynes:
I would but, alas, I am no user of TypeScript and so cannot pass judgement on the merits of this PR.

rhysd:
@nhynes 

Hmm... So currently this repo doesn't have a reviewer regarding to TypeScript.  I think I can add a test for .d.ts file.  We can check the test result for PR related to TypeScript.

PR #20, opened 2016-08-17
Fix broken TypeScript type definition and introduce test 

rhysd:
I added a test for TypeScript type definitions and integrated it into CI.

And I fixed some broken points in `index.d.ts`

- `declare module` should not be used for type definitions resolved as npm package
- `Nvim` should inherit `EventEmitter`
- Add missing `equals()` methods


justinmk:
Thanks!

PR #21, opened 2016-08-28
Add promise API

billyvg:
This includes a new `promise` module in the package. 

Reference:
#7 
https://github.com/neovim/node-host/issues/15


billyvg:
Looks like tests are passing for node 0.12, but failing on node 0.10 because Promises were not added yet in 0.10

rhysd:
You can use polyfill such as [es6-promise](https://www.npmjs.com/package/es6-promise).  But node.js v0.10 is too old and I think it should no longer be supported.

nhynes:
No polyfills, please! I don't think that that v0.10 is "too old," however, considering that
* its last release was two months ago and
* [as a heuristic] all but the latest release of ubuntu use some flavor of v0.10.

My humble recommendation is to either merge this into master but not cut a release or place a guard on the entire promise module that checks if the required features are present. My preference is the former.

Delapouite:
I'm in favor of dropping support for v0.10 : the end of maintenance is in October. And you can easily install a newer versions https://github.com/creationix/nvm

nhynes:
@Delapouite cool, thanks for the link and your feedback. Let's, then, keep this project in line with the progressive nature of neovim, itself, and only support >= 0.12.

Looking forward to merging this today (using the 24 hour definition, rather than whatever one I used earlier... :)

nhynes:
Okay, the tests are passing again. Looking at this PR a bit more closely, it seems like you have to explicitly import the `neovim-client/promise` module to use promises. Is there any reason to prefer this to automatically returning a Promise when a callback isn't provided as has been proposed in `node-host`?

Otherwise, the PR, once documented, looks ready to go!

rhysd:
@nhynes 

Please see this comment: https://github.com/neovim/node-host/issues/15#issuecomment-238096773 .

Promise API requires 'request' rather than 'notification' because API can't know user will use the response or not.

nhynes:
@rhysd good point, thanks; I was under the impression . maybe someone should write up a static analysis tool, real quick, to tell if the promise is ever used later ;)

Anyway, I'll finish updating the readme.

nhynes:
Okay, everything seems to be working now. There was a tiny bug in which methods without callbacks were being promisifed, but that's been fixed.

Thanks again for the PR!

PR #22, opened 2016-09-04
Use 'stable' and 'lts/*' to reduce maintain cost for Travis CI

rhysd:
Currently Node.js has two primary branches:

- stable: latest stable release
- LTS: long-term supported release

I think specifying them is better than specifying versions directly. We no longer need to take care about Node.js versions on Travis CI.


nhynes:
Thanks!

PR #23, opened 2016-11-06
Generate new typescript defs for 0.1.6

davecoates:
Also includes definitions for promise.d.ts

Fixes incorrect type definitions for some methods (eg. bufGetNumber)
that the API doesn't return argument types for - I presume this is
because these methods aren't implemented. Previously they got same
type signature as callback; now it will be `any`.


justinmk:
Thanks. cc @billyvg @rhysd @nhynes

billyvg:
Closed for #28

PR #25, opened 2016-12-26
Skip deprecated APIs and fix parameter types of non-method functions (Fix #24)

### LINKED ISSUE #24: Incorrect type definitions
Hi,

Thanks for the awesome library! However, I'm having a slight problem: some of the type definitions seem to be incorrect. For instance, with this one:

```ts
getLine(index: Buffer, cb: number): void;
```

It seems like the class type (a.k.a. `Buffer`) got prepended to the arguments list, which put everything one place to the right. Am I correct? If so, is there any chance this could be fixed? It's not a major problem, but I'm currently using the type definitions as a reference, which make finding the right args just a littlebit more difficult.

rhysd:
I fixed below points

- `vim_*` functions are deprecated and replaced with `nvim_*`. So we don't need to generate RPC functions for them.
- Condition for shifting parameters of non-method functions was wrong
- Tests failed because nvim 0.2 changed some API names

But tests are still failing. `nvim_set_current_win` looks not working (I could reproduce this also in my local environment). `nvim_set_current_buf` is working (weird...).

CC: @justinmk 

- Environment
  - nvim 0.2.0
  - OS X 10.11.6

```
NVIM v0.2.0-164-gc1d1493
Build type: RelWithDebInfo
Compilation: /usr/local/Homebrew/Library/Homebrew/shims/super/clang -Wconversion -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -O2 -g -DDISABLE_LOG -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wvla -fstack-protector-strong -fdiagnostics-color=auto -DINCLUDE_GENERATED_DECLARATIONS -DHAVE_CONFIG_H -I/tmp/neovim-20161226-8685-1cx1y49/build/config -I/tmp/neovim-20161226-8685-1cx1y49/src -I/usr/local/include -I/usr/local/include -I/usr/local/include-I/usr/local/include -I/usr/local/include -I/usr/local/include -I/usr/local/opt/gettext/include -I/usr/include -I/usr/include -I/tmp/neovim-20161226-8685-1cx1y49/build/src/nvim/auto -I/tmp/neovim-20161226-8685-1cx1y49/build/include
Compiled by npsdev5@OfficeDev.local

Optional features included (+) or not (-): +acl   +iconv    +jemalloc +tui
```


justinmk:
> nvim_set_current_win looks not working (I could reproduce this also in my local environment). 

@rhysd It expects a window "id", not a "number". That is, the result of VimL function `win_getid()`.

rhysd:
@justinmk 

As long as I tried, the type of `window` paramerter is `Window`, not an `Integer`.

```json
{"parameters":[["Window","window"]],"method":false,"return_type":"void","name":"nvim_set_current_win","since":1}
```

Is this correct?

justinmk:
@rhysd The Window msgpack type is an alias for integer.

test/index.js
```diff
@@ -71,19 +72,27 @@ describe('Nvim', function() {
...
-      nvim.getWindows(function(err, windows) {
+      strictEqual(err, null);
+      nvim.listWins(function(err, windows) {
+        strictEqual(err, null);
         equal(windows.length, 2);
         equal(windows[0] instanceof nvim.Window, true);
         equal(windows[1] instanceof nvim.Window, true);
-        nvim.setCurrentWindow(windows[1], function(err, res) {
-          nvim.getCurrentWindow(function(err, win) {
+        nvim.setCurrentWin(windows[1], function(err, res) {
```
justinmk: Is `nvim.listWins` returning the wrong thing?

justinmk: Just as a sanity check, does this pass?

    nvim.setCurrentWin(0, function(err, res) {

rhysd:
@justinmk 

Oh, I didn't know that... How about other types (`Buffer` or `Tabpage`)? I don't know how to get which type should be object and which should be an alias of primitive types.

test/index.js
rhysd: Thanks. I'll try tomorrow.

justinmk:
@rhysd Buffer, Tabpage, and Window are all aliases to integer.

justinmk:
@nhynes Are you available to help with this? I'm trying my best not to install nodejs.

nhynes:
> Are you available to help with this?

Heh, same. I can try to take a look at this over this weekend, but if it will require significant debugging, it will, regrettably, have to wait until next weekend (but hopefully before then!).

tjdevries:
I started some work at https://github.com/tjdevries/node-client but have not finished it yet (thus no PR), to try and do several of the things mentioned in this PR. It's still pretty WIP.

My  goal was to rewrite most of the generation in typescript, but also keep the transpiled javascript in the repo as well, so typescript is not a requirement (you'll find `.ts` files in `src/`  and `.js` files in `lib/`).

rhysd:
> Buffer, Tabpage, and Window are all aliases to integer.

Oh... I see. Currently `Tabpage` and `Window` and `Buffer` are implemented as class and many APIs are implemented as their methods.
However, if we change them to integer, we no longer can`t make them class. I think this would need large code changes and API changes.

test/index.js
justinmk: So did this work ?

tjdevries:
I think you should have the parameters be "BufferHandle" or something like that, and the results of those things can still be "Buffer" which have the methods.

billyvg:
Closed by #28

PR #27, opened 2017-01-04
[WIP] Modernize codebase using TypeScript and promises

samvv:
If interested, this is related to tarruda/node-msgpack5rpc#6 but now for the Neovim client.

Some of the changes:
 - No use of eval: instead, just create a new closured lambda. The only disadvantage I could find is that `Function.length` and `Function.name` does not return the correct value.
 - Make every method return a promise by default

This pull introduces some breaking changes so it is best to bump the version number to `2.0.0` you think it is worthwhile to merge.

**Note:** I have not upgraded the version number of `msgpack5rpc` in `package.json`. This has to be done manually.


tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "compilerOptions": {
+    "target": "es3",
```
rhysd: This module is cosidered to use with Node.js. So I want to suggest `"es5"` as a target.

typings.json
```diff
@@ -0,0 +1,5 @@
+{
+  "globalDependencies": {
+    "mocha": "registry:dt/mocha#2.2.5+20161221010438"
```
rhysd: Instead of using typings, I think `npm i -D @types/mocha` is better.

index.ts
```diff
@@ -0,0 +1,143 @@
...
+
+        this._channel_id = channel_id
+        this._types = types
+
+        const sessionTypes = _.map(types, ({ version, id, prefix }, name) => {
+
+          class NvimType implements NvimType {
+
+            _data: Buffer
+            _client: NvimClient
```
rhysd: Should these members be private?

index.ts
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  private generateWrappers(functions) {
+
+    for (const func of functions) {
+
+      const className = _.capitalize(func.name.substring(0, func.name.indexOf('_')))
+      const params = func.parameters.map(([type, name]) => ({ type, name }))
+      const methodName = _.camelCase(func.name.substring(func.name.indexOf('_')))
+      const isGlobal = ['Vim', 'Nvim', 'Ui'].indexOf(className) !== -1
```
rhysd: `vim_*` are deprecated APIs. We might need to ignore them.

samvv:
Thanks for the feedback, I'll get to them in a couple of days.

justinmk:
Wasn't node-client already "promisified"? https://github.com/neovim/node-client/commit/a1dd8a6e8def020e0f831c01f8a720c654d73409

@tarruda Makes good points in https://github.com/tarruda/node-msgpack5rpc/pull/6 which may apply here as well. In particular, what's the benefit? In general, we want to avoid breaking changes.

samvv:
I'll admit it was more of an exercise to get to know the codebase, and I get your point. Closing this.

PR #28, opened 2017-06-16
[Breaking changes] New client/plugin/host APIs

billyvg:
This change would break backwards compatibility, but I wanted to see if this was a direction people would be interested in moving towards. If not I can just maintain a different fork. Here's a summary of some changes:

* Moved `node-host` support into this package
* APIs are all promises now
* API wrappers that are modeled after `python-client`. i.e. to get current buffer name you can do this:
```javascript
const bufferName = await nvim.buffer.name;
```
* Will move towards removing ALL dynamically generated APIs. This will mean there will be a stable API.
* Modernized codebase + will be moving towards Typescript. Currently only tested on node 7-8, but will be able to support 6 once ported to TS. 
* Remote plugin API currently relies on decorators (open to other ideas though, on this and everything else)
* More and better ways to debug


justinmk:
@billyvg Thanks, most of those bullet points sound very nice.

> Moved node-host support into this package

Generally I'm in favor of reducing repo fragmentation. We have python host in the python-client and it feels right. 

> This change would break backwards compatibility,

I guess we could bump the npm package version to indicate breaking change. And the old version will continue to be available on npm, right?

> APIs are all promises now

@tarruda objected to this for what seemed to me a good reason. But people keep pushing for it. Isn't it trivial to promisify a non-promisified js API?

Would appreciate input from @rhysd @nhynes @snoe @mhartington @chillee .

Would like to bring in more people to maintain this repo...

Chillee:
@justinmk  Sure it's possible, but Bluebird is just another dependency and returns non-native promises. The non-native promises is a big one, as those promises need to be handled separately, and just cause more technical debt within our codebase.

Promises are the asynchronous solution that the javascript community has solidified around, and I'd prefer it.

tarruda:
> @tarruda objected to this for what seemed to me a good reason. But people keep pushing for it. Isn't it trivial to promisify a non-promisified js API?

Yes.  [Bluebird.promisify](http://bluebirdjs.com/docs/api/promise.promisify.html) essentially converts node-style callback-based APIs into promise-based APIs. Many packages in node ecosystem have promisified versions that simply call this function.

mhartington:
Big 👍  for this. 

> tarruda objected to this for what seemed to me a good reason. But people keep pushing for it. Isn't it trivial to promisify a non-promisified js API?

Yep, it's pretty easy to create a new Promise from a non-promise API. 
If there's a question of API support,  Node has support all the back to 6x, which is in LTS now.

http://node.green/#ES2015-built-ins-Promise

There _could_ be hard requirement on minimum node version

>  Currently only tested on node 7-8, but will be able to support 6 once ported to TS.
As for this, it's mostly because the client will make use of async/await internally. 
Merging could be held off until I wrap up the typescript port.

tarruda:
> The non-native promises is a big one, as those promises need to be handled separately, and just cause more technical debt within our codebase.

Can you elaborate on this argument? AFAIK there's nothing special about native promises and you can use them interchangeably with other promises, as long as the API is the same.

In fact, V8 native Promise class is implemented in pure javascript, and Bluebird Promises actually has a faster implementation([ref](https://softwareengineering.stackexchange.com/questions/278778/why-are-native-es6-promises-slower-and-more-memory-intensive-than-bluebird), not that peformance should be a significant factor in this decision, just thought it would be important to mention this to demystify native Promises.).

Chillee:
@tarruda Yeah you're right. I remembered incorrectly that certain syntaxes (await was what I was thinking of) didn't work with Bluebird Promises.

They seem to be interchangeable, so I have no objection to using Bluebird. I think still it's a good idea for this package to provide Promises (whether using Bluebird or something else).

billyvg:
My initial reasonings for being a Promise only library was:
* rewriting node-host + client meant breaking changes, so this was an opportunity to modernize the codebase. I figured I'd be able to move quicker and made the code cleaner with Promises.
* node 4.x is reaching EOL next year and node 6.x has native support for Promises.

I think it's better to use the more modern approach to async handling and have plugins/dependents do the same.  From the looks of things, there are not many remote node plugins -- the most common use-case of the current node client is for UI clients (i.e. NyaoVim, oni, VSCodeVim) and all three use Promises. Why keep support for callbacks?

justinmk:
Yeah, we don't have many users of node-client because its bugs aren't being fixed. So if this leads to a project that is actively maintained then IMO I don't care how it happens :)

justinmk:
I invited several people as maintainers of this repo and node-host, with the hope that @billyvg will take the lead.

bin/generate-typescript-interfaces.js
```diff
@@ -0,0 +1,93 @@
...
+  Object.keys(interfaces).forEach(key => {
+    let name = key;
+    if (key === 'Neovim') {
+      name += ' extends NodeJS.EventEmitter';
+    }
+    process.stdout.write(`export interface ${name} {\n`);
+    if (key === 'Neovim') {
+      process.stdout.write('  quit(): void;\n');
+      process.stdout.write('  isApiReady(): Boolean;\n');
+      process.stdout.write('  requestApi(): Promise<[integer, any]>;\n');
```
Chillee: I think this should be `number` right?

bin/generate-typescript-interfaces.js
```diff
@@ -0,0 +1,93 @@
...
+    Buffer: nvim.Buffer,
+    Window: nvim.Window,
+    Tabpage: nvim.Tabpage,
+  };
+
+  // use a similar reference path to other definitely typed declarations
+  process.stdout.write('interface AttachOptions {\n');
+  process.stdout.write('  writer?: NodeJS.WritableStream,\n');
+  process.stdout.write('  reader?: NodeJS.ReadableStream,\n');
+  process.stdout.write('  proc?: NodeJS.ChildProcess,\n');
```
Chillee: Should this be NodeJs.Process?

billyvg:
Merging this into the `next` branch of this repo. Will create a new PR when we're ready to merge this to `master`.

PR #34, opened 2017-10-26
Updated typings for atomic call

Chillee:
The typing for atomic call should be any[].

### DIFF
--- a/src/api/Neovim.ts
+++ b/src/api/Neovim.ts
@@ -161,7 +161,7 @@ export class Neovim extends BaseApi {
   }
 
   /** Call Atomic calls */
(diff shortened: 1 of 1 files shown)


src/api/Neovim.ts
```diff
@@ -161,7 +161,7 @@ export class Neovim extends BaseApi {
   }
 
   /** Call Atomic calls */
-  callAtomic(calls: Array<string>): Promise<[Array<any>, boolean]> {
+  callAtomic(calls: Array<any>): Promise<[Array<any>, boolean]> {
```
billyvg: Would a more specific type be `VimValue`?

Chillee: I think that would probably be better.

PR #35, opened 2017-10-26
Move typescript package to devlopment dependency

rhysd:
If my understanding is correct, TypeScript is only used in development environment. So I moved the package from `dependencies` to `devDependencies` to reduce runtime dependencies.


billyvg:
Thanks!

PR #36, opened 2017-10-27
Fix TypeScript errors and test hang problem

rhysd:
I fixed following two points:

1. `tsc` errors. I used `new Logger` to make a logger instance also for console output in `utils/logger.ts`.
2. I noticed that test hung when `neovim-node-host` command was not found. I fixed it by adding check in `autoload/provider/node.vim`

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+package-lock=false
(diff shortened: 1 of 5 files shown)


rhysd:
work conflict... I'll create PR again considering the latest changes.

Issue #47, opened 2017-12-18
Run tests on Appveyor to debug rplugin on Windows


janlazo:
After work done on https://github.com/neovim/neovim/pull/7706, https://github.com/neovim/neovim/pull/7734, https://github.com/neovim/neovim/pull/7738, I can't make a simple echo class with an `EchoHello` command work on Windows. The APIs are fine because testing on them directly, bypassing the host and `rplugin/node`,  works on all OS consistently. `:checkhealth` and `:UpdateRemotePlugins` are ok now (on Windows) and I see `rplugin.vim` with `EchoHello` registered from nodejs plugin I made it.

That leaves msgpack and the `RPC` api from both sides which I don't know how to debug/fix yet.
To get more debug info, the tests should run on Appveyor as well. We can reuse the same configuration from neovim repo.

janlazo:
NVIM_NODE_LOG_FILE

```
2017-12-20T18:15:31.532Z - debug: host.start

2017-12-20T18:16:06.580Z - debug: host.start

2017-12-20T18:16:06.612Z - info: handleRequest:  poll

2017-12-20T18:16:06.626Z - debug: request received:  poll

2017-12-20T18:16:06.629Z - info: handleNotification:  C:/Users/JanEdmund/AppData/Local/nvim/rplugin/node/hello.js:command:EchoHello

2017-12-20T18:16:06.630Z - debug: host.handlePlugin:  C:/Users/JanEdmund/AppData/Local/nvim/rplugin/node/hello.js:command:EchoHello

2017-12-20T18:16:06.766Z - info: [C] Error loading child ChildPlugin C

2017-12-20T18:16:06.789Z - info: [C] { Error: Cannot find module 'C'
    at Function.Module._resolveFilename (module.js:469:15)
    at Function.Module._load (module.js:417:25)
    at Module.require (module.js:497:17)
    at Object.sandboxRequire [as require] (C:\Users\JanEdmund\lib\node6\node_modules\neovim\lib\host\factory.js:79:46)
    at createPlugin (C:\Users\JanEdmund\lib\node6\node_modules\neovim\lib\host\factory.js:105:39)
    at Object.loadPlugin (C:\Users\JanEdmund\lib\node6\node_modules\neovim\lib\host\factory.js:139:16)
    at Host.getPlugin (C:\Users\JanEdmund\lib\node6\node_modules\neovim\lib\host\index.js:23:59)
[...]

PR #48, opened 2017-12-20
host: handle network drive for Windows filename

janlazo:
Fix https://github.com/neovim/node-client/issues/47#issuecomment-353062143

### DIFF
--- a/src/host/index.ts
+++ b/src/host/index.ts
@@ -36,7 +36,11 @@ export class Host {
       logger.debug('host.handlePlugin: ', method);
 
       // Parse method name
-      const procInfo = method.split(':');
+      let procInfo = method.split(':');
+      if (process.platform === 'win32') {
+        const networkDrive = procInfo.shift();
+        procInfo[0] = networkDrive + ':' + procInfo[0];
+      }
       const filename = procInfo[0];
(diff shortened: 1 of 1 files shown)


janlazo:
Travis failure seems unrelated to the patch (use `let` to modify a value if running on Windows).
`stable` is an alias to `node` so nvm installs 9.3.0, latest stable version of node.
It should test on LTS versions (6.x.x, 8.x.x) instead.

janlazo:
@billyvg Can you review my changes? I don't encounter any issues on node 6.12.x with my changes.

src/host/index.ts
```diff
@@ -36,7 +36,11 @@ export class Host {
       logger.debug('host.handlePlugin: ', method);
 
       // Parse method name
-      const procInfo = method.split(':');
+      let procInfo = method.split(':');
+      if (process.platform === 'win32') {
+        const networkDrive = procInfo.shift();
```
billyvg: Can you add a comment with example `procInfo` on windows?

janlazo:
The integration tests should pass with #50 now.

rhysd:
I'm merging this because CI failure is not related to changes of this PR. Thank you @janlazo

PR #50, opened 2017-12-24
Fix integration tests on Travis

janlazo:
Debug CI issues caused by https://github.com/neovim/neovim/pull/7734, https://github.com/neovim/neovim/pull/7738, and any issues with the latest stable version of nodejs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,7 @@ script:
   - export NVIM_NODE_LOG_LEVEL=debug
   - yarn
   - yarn link
+  - npm install -g ./
   - yarn run prep-integration-test
   - yarn test -- --coverage
 


janlazo:
Since the test passed, can the integration tests use `NODE_PATH` and `require('neovim')` similar to https://github.com/neovim/neovim/blob/master/test/functional/provider/nodejs_spec.lua?

.travis.yml
```diff
@@ -28,6 +28,7 @@ script:
   - export NVIM_NODE_LOG_LEVEL=debug
   - yarn
   - yarn link
+  - npm install -g ./
```
rhysd: Thank you for your fix. I wonder why this fixes CI because `yarn link` links this repository to the global. Could you please explain the reason?

janlazo: `npm link` adds the `neovim` module in the folder output of `npm root -g`.
`yarn link` doesn't.
`provider#node#Detect()` on neovim HEAD directly checks if `$(npm root -g)/neovim/bin/cli.js` exists.

rhysd: Thank you for your explanation. I think it should be improved. But it is not a scope of this PR.

PR #52, opened 2018-03-04
Fixed Plugin TypeScript annotations, to keep the TypeScript compiler …

FredrikAleksander:
…from failing when using the Plugin decorator in TypeScript

When using the Neovim module from TypeScript, the TypeScript compiler fails when using the Plugin decorator because the signature is not what it expects from a decorator. The fix adds proper annotations to the function declaration. The return value has to be cast, as the circular types PluginWrapperInterface and PluginWrapperConstructor will trip out the compiler. Added a test case, but this currently passes even without the fix, as the tests are not type checked according to tsconfig.json (the type checking is what is failing).

### DIFF
--- a/src/plugin/plugin.ts
+++ b/src/plugin/plugin.ts
@@ -35,7 +35,7 @@ function wrapper(
 }
 
 // Can decorate a class with options object
-export function plugin(outter: any) {
+export function plugin<T extends { new(nvim: Neovim): {}}>(outter: any): (constructor: T) => T {
   /**
     * Decorator should support
     *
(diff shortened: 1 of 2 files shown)


src/plugin/plugin.ts
```diff
@@ -35,7 +35,7 @@ function wrapper(
 }
 
 // Can decorate a class with options object
-export function plugin(outter: any) {
+export function plugin<T extends { new(nvim: Neovim): {}}>(outter: any): (constructor: T) => T {
```
justinmk: might as well fix the spelling of `outer` now :)

justinmk:
@FredrikAleksander can you include the explanation in the commit message?

FredrikAleksander:
Fixed the typo and updated the commit message

justinmk:
Not sure how to publish a new release, deferring  to @billyvg for merge.

FredrikAleksander:
Having worked some more on a plugin locally with the changes, I've found some more issues. And I dont think they are very easily solved as is. So this change by itself would not be enough by itself. I've managed to write a simple module that just uses the other parts the library, and implements it's own decorators that I use instead. So this pull request isn't really needed in the end. I see there is an open issue regarding moving decorators, and I might drop some comments there.

Issue #43, opened 2017-11-20
Reduce complexity of the plugin API 


kutsan:
I think API should be vanilla NodeJS friendly. Babel and friends don't help developers at this point and using legacy decorators and other early features shouldn't be recommended. Please keep the API simple.

I could write the plugins like this now without Babel configurations but decorators looks bad to be honest.

```javascript
const { Plugin, Command } = require('neovim')

class TestPlugin {
	splitV() {
		this.nvim.command('vsplit')
	}
}

Command('Vsplit')(TestPlugin.prototype.splitV)

module.exports = Plugin({ dev: true })(TestPlugin)
```

billyvg:
I agree, I'll explore a simpler API for creating plugins

kutsan:
Thanks for your answer. I'm looking forward to.

janlazo:
Doesn't work in Windows. I've tried on node 6.91 and current LTS (8.9.1).

kutsan:
@janlazo I wish I could help but I don't use and know Windows, but it should work; I don't think the problem this about the code sample. For what it's worth, that's not this issue for. I think you should create an another issue for that, with your exact steps.

janlazo:
This works for me in Windows. I can't register commands yet.

```javascript
const socket = process.env.NVIM_LISTEN_ADDRESS;
const neovim = require('neovim');
const nvim = neovim.attach({socket: socket});

class TestPlugin {
    echoHello() {
        this.nvim.command('echom "hello"')
    }
}

neovim.Command('Hello')(TestPlugin.prototype.echoHello);
const PluginClass = neovim.Plugin(TestPlugin);
const plugin = new PluginClass(nvim);
plugin.echoHello();
```

justinmk:
Is there a way to avoid users/plugin authors needing to manually manage the `$NODE_PATH` ?

In the python-client we mangle the import path dynamically.  Although virtualenv's must be managed by the user (obviously).

billyvg:
@janlazo You shouldn't need to do `jobstart(['node', 'path/to/plugin.js'])`, that's what the node host is supposed to handle.

janlazo:
@billyvg What about plugins outside of `rplugin/node`?

billyvg:
What's the use case where it's outside of `rplugin`?

billyvg:
How do you feel about something like this:

```javascript
const NeovimPlugin = require('neovim/plugin');

[...]

PR #54, opened 2018-03-13
Functional api - Implements #43

simlrh:
An alternative version of the plugin API based on my suggestions in #43. The API remains backwards compatible with the decorator API.

See the [README](https://github.com/simlrh/node-client/blob/feature/alternative-api/README.md) or the test cases for examples of how the API is used. Briefly it's something like:

```js
export default (plugin) => {
  plugin.registerCommand('CommandName', () => plugin.nvim.setLine('CommandName called!'));
}
```


README.md
```diff
@@ -59,28 +59,130 @@ const nvim_proc = cp.spawn('nvim', ['-u', 'NONE', '-N', '--embed'], {});
 ```
 
 ## Writing a Plugin
+If you are a plugin developer, I'd love to hear your feedback on the plugin API.
+
 A plugin can either be a file or folder in the `rplugin/node` directory. If the plugin is a folder, the `main` script from `package.json` will be loaded.
 
+The plugin should export a function which takes a `NeovimPlugin` object as it's only parameter. You may then register autocmds, commands and functions by calling methods on the `NeovimPlugin` object. You should not do any heavy initialisation or start any async functions at this stage, as nvim may only be collecting information about your plugin without wishing to actually use it. You should wait for one of your autocmds, commands or functions to be called before starting any processing.
```
justinmk: Technical identifiers should always use "nvim", not "neovim".  E.g. in this case, `NvimPlugin`, not `NeovimPlugin`.

Any namespaces or whatever that spell "neovim" are historical accidents.

justinmk:
(Those codecov comments are annoying, need to figure out how to remove them.)

@simlrh This looks like a great effort, very appreciated.

I don't keep up with JS-land so I'm completely lost about why people constantly want to overhaul the node-client API.

Did you see these previous discussions? (Some of the discussions are on the old `node-host` project, which was later merged-into `node-client`)

- https://github.com/neovim/node-host/issues/15
- [Promise API](https://github.com/neovim/node-host/pull/16) (by @billyvg , who AFAIK did an overhaul of the API a few months ago)
- [Modernize codebase using TypeScript and promises](https://github.com/neovim/node-client/pull/27)

Are those discussions relevant to this "Functional api" change? What are your thoughts?

I hope @neovim/nodejs people can review this.

simlrh:
@justinmk Thanks, I will change those variable names in the next couple of days.

I believe the linked discussions mostly refer to the nvim api object which JS plugins use to call nvim functions etc, whereas this PR addresses the way JS plugins register their own commands and handle requests from nvim.

I actually really like the current decorator based api, but decorators are a stage 2 draft proposal and not yet a part of standard JavaScript, and they are not supported by any version of node, so a transpilation build step and associated tooling is necessary for all plugins. This PR still allows that behaviour, but you can also simply write one .js file which sets function callbacks for your commands. 

Another goal was to make it possible to dynamically add and remove listeners for arbitrary rpcnotify events. It’s not implemented in this PR but it would only require adding another method to NvimPlugin.

mhartington:
Great PR! Will pull down and test it out!

billyvg:
Thanks @simlrh! It's a large diff so it might take me a bit to review but it looks great at a glance!

SevereOverfl0w:
I installed this locally using `npm link`, and I cannot get it to run:

```
 neovim-node-host --version
module.js:545
    throw err;
    ^
Error: Cannot find module '../lib/host'
    at Function.Module._resolveFilename (module.js:543:15)
    at Function.Module._load (module.js:470:25)
    at Module.require (module.js:593:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/home/dominic/src/github.com/simlrh/node-client/bin/cli.js:2:14)
    at Module._compile (module.js:649:30)
    at Object.Module._extensions..js (module.js:660:10)
    at Module.load (module.js:561:32)
    at tryModuleLoad (module.js:501:12)
    at Function.Module._load (module.js:493:3)
```

Issue #58, opened 2018-05-04
Bug:  expand autocmd with <abuf> is wrong for BufUnload


chemzqm:
```
  @Autocmd('BufUnload', {
    sync: true,
    pattern: '*',
    eval: 'expand("<abuf>")'
  })
  public async onBufUnload(bufnr: string):Promise<void> {
    logger.debug(`file: ${bufnr} remove`)
  }
```
The `bufnr` here is always the buffer number of current buffer number, but it should be the buffer number of removed buffer.
```
autocmd BufUnload * call s:OnUnload()
function! s:OnUnload()
  echo +expand('<abuf>')
endfunction
```
vim script doesn't have this problem.

BTW: `<afile>` is neither correct.

PR #67, opened 2018-05-13
fix: onBufUnload gives wrong buffer id (WIP)

billyvg:
This is currently a WIP to address #58, I've only added a failing test case for this behavior. Someone feel free to pick this up because I'm not sure what's wrong.


chemzqm:
I think it's bug of neovim host, it's running expand expression async.

PR #77, opened 2018-05-19
feat(client): add support for buffer notifications

chemzqm:
Make node-client works with PR https://github.com/neovim/neovim/pull/7917

### DIFF
--- a/src/api/Buffer.ts
+++ b/src/api/Buffer.ts
@@ -112,6 +112,16 @@ export class Buffer extends BaseApi {
     });
   }
 
+  /* Turn on live updates */
+  attach(send_buffer: boolean) {
+    return this.request(`${this.prefix}attach`, [this, send_buffer]);
+  }
+
+  /* Turn off live updates */
+  detach() {
+    return this.request(`${this.prefix}detach`, [this]);
+  }
+
   /** Get buffer name */
   get name(): string | Promise<string> {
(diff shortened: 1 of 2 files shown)


src/host/index.ts
```diff
@@ -33,6 +33,7 @@ export class Host {
 
   // Route incoming request to a plugin
   async handlePlugin(method: string, args: any[]) {
+    if (method.startsWith('nvim_buf_')) return
```
justinmk: Just for reference, our plan is to reserve the `nvim_` prefix for events. Also, all events will have the `_event` suffix.

So maybe it makes sense to ignore `nvim_`, not just `nvim_buf_` ?

chemzqm: Sure.

src/api/Buffer.ts
```diff
@@ -112,6 +112,16 @@ export class Buffer extends BaseApi {
     });
   }
 
+  /* Turn on live updates */
+  attach(send_buffer: boolean) {
```
billyvg: Can you change `send_buffer` to be `sendBuffer`?

billyvg:
I just read up on https://github.com/neovim/neovim/pull/7917#issuecomment-392744195 - I'm going to change the interface for this and remove `attach`/`detach` here (`attach` also doesn't work, missing an arg).

PR #79, opened 2018-06-01
fix(): slow performance on decode large data

chemzqm:
`msgpack-lite` and `msgpack5` could be quite slow when working with streams that emit small chunks (8kb) of data each time,  it could be solved by using a buffered stream in the middle.

Detail at https://github.com/chemzqm/stream-issue

### DIFF
--- a/src/utils/buffered.ts
+++ b/src/utils/buffered.ts
@@ -1,0 +1,55 @@
+import { Transform } from 'stream';
+
+const MIN_SIZE = 8 * 1024;
+
+export default class Buffered extends Transform {
+  private chunks: Buffer[] | null;
+  private timer: number | null;
+  constructor() {
+    super({
+      readableHighWaterMark: 10 * 1024 * 1024,
+      writableHighWaterMark: 10 * 1024 * 1024,
+    } as any);
+    this.chunks = null;
+    this.timer = null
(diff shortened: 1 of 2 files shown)


src/utils/buffered.ts
```diff
@@ -0,0 +1,55 @@
...
+      this.sendData();
+      callback();
+      return;
+    }
+    if (!chunks) {
+      chunks = this.chunks = [chunk];
+    } else {
+      chunks.push(chunk);
+    }
+    this.timer = setTimeout(this.sendData.bind(this), 100);
```
billyvg: Do we need the 100ms timeout? Would it work if we just delayed `sendData` to the next execution cycle (e.g. with 1ms timeout)?

There's an edgecase if chunk size is exactly a multiple of `8*1024`, then `sendData` will have an added 100ms delay to it.

chemzqm: I'm tried with 1ms and 10ms timeout, it doesn't work most of times, changed to 20ms instead.

billyvg: OK, we may want to change the buffering strategy or adopt a buffered stream dependency, but I think this is a big improvement and worthwhile compromise. Thanks!

PR #87, opened 2018-06-11
feat(ci): Allow for missing APIs failure

billyvg:


rhysd:
Yeah, I agree CI should not fail when Neovim is updated and missing APIs are detected because it will block other PRs. However, I concern that no one does not notice a CI failure. Do you have any ideas on this?

PR #89, opened 2018-06-15
feat(api): add id attribute to Buffer and Window

chemzqm:
The `data` of Buffer and Window would always be the buffer number and the number of windowid. (decoded by messagepack-lite).

This PR makes plugin have associate data with buffer and window easier.

Also fix jest not working on latest node.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "winston": "^2.3.1"
   },
   "devDependencies": {
-    "@types/jest": "^20.0.1",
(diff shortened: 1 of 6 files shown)


megalithic:
@billyvg any ideas or timelines around a new version of the node-client getting published? this specific PR explicitly enables https://github.com/neoclide/coc.nvim/issues/2 to begin working.

billyvg:
@megalithic just released 4.2.0

PR #91, opened 2018-07-14
fix(api): reattach of buffer sometimes not working

chemzqm:
Mark `isAttached` to false on detach.

Note that there is a wait in the test, since attach again without detach returned would take no effect. May be the detach should return a promise to help this case.


billyvg:
Thanks, published as `4.2.1`

PR #93, opened 2018-08-18
fix(buffer): fix attach events

chemzqm:
* Not use internal isAttached
* Add buffer.isAttached method
* Unbind listener if attach failed

### DIFF
--- a/src/api/Buffer.ts
+++ b/src/api/Buffer.ts
@@ -24,7 +24,10 @@ export const ATTACH = Symbol('attachBuffer');
 
 export class Buffer extends BaseApi {
   public prefix: string = Metadata[ExtType.Buffer].prefix;
-  private isAttached: boolean = false;
+
+  public get isAttached(): boolean {
+    return this.client.isAttached(this);
+  }
 
   /**
    * Attach to buffer to listen to buffer events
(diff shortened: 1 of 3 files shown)


justinmk:
ping @neovim/nodejs

src/api/Buffer.ts
```diff
@@ -33,8 +36,17 @@ export class Buffer extends BaseApi {
...
    *        a `nvim_buf_changedtick_event`
    */
-  [ATTACH] = (sendBuffer: boolean = false, options: {} = {}) =>
-    this.request(`${this.prefix}attach`, [this, sendBuffer, options]);
+  [ATTACH] = async (
+    sendBuffer: boolean = false,
+    options: {} = {}
+  ): Promise<boolean> => {
+    if (this.client.isAttached(this)) return true;
+    return this.request(`${this.prefix}attach`, [
```
billyvg: Let's use `await` here:

```javascript
try {
 return await this.request(`${this.prefix}attach`, [this,
      sendBuffer,
      options,
    ]);
} catch(err) {
 return false;
}
```

src/api/Buffer.ts
```diff
@@ -243,20 +255,22 @@ export class Buffer extends BaseApi {
    */
   listen(eventName: string, cb: Function): Function {
     if (!this.isAttached) {
-      this[ATTACH]();
-      this.isAttached = true;
+      this[ATTACH]().then(attached => {
```
billyvg: `await` instead of `then` here too (will also have to make this method `async`)

chemzqm: I'm fine with `async`, but that would be break change.

billyvg: Ah you're right, I didn't read down far enough, let's revert this change. Sorry about that.

PR #95, opened 2018-10-01
[RFC] debug: spawn child process with --inspect-brk

### LINKED ISSUE #46: $NVIM_NODE_HOST_DEBUG should not rely on the end user running cli.js directly
See https://github.com/neovim/neovim/pull/7706#issuecomment-351463151

I can't add `--inspect-brk` on node.exe on Windows because of how the npm shims work. On Ubuntu, `neovim-node-host` is a javascript shim (with `/usr/bin/env node` as first line) so the user can run it with `node` directly instead of relying on the shell to do it for you. That doesn't work in Windows, even on Cygwin, so the only way to add it is something like `node --inspect-brk C:\node\node_modules\neovim\bin\cli.js`, ignoring the npm shims.

I prefer that this be resolved on `bin/cli.js` by restarting itself with `--inspect-brk` if the environment variable is defined.

janlazo:
Fix #46 

I included style changes because of the precommit check with husky.
Running node with `--inspect-brk` silences the following error whether I spawn a new host or not:

```
net.js:704
    throw new TypeError(
    ^

TypeError: Invalid data, chunk must be a string or buffer, not object
    at WriteStream.Socket.write (net.js:704:11)
    at Transport.parseMessage (/home/janlazo/repo/git/node-client/lib/utils/transport.js:105:31)
    at DecodeStream.Transport.decodeStream.on (/home/janlazo/repo/git/node-client/lib/utils/transport.js:37:18)
    at emitOne (events.js:116:13)
    at DecodeStream.emit (events.js:211:7)
    at addChunk (_stream_readable.js:263:12)
    at readableAddChunk (_stream_readable.js:250:11)
    at DecodeStream.Readable.push (_stream_readable.js:208:10)
    at DecodeStream.Transform.push (_stream_transform.js:147:32)
    at DecodeBuffer.DecodeStream.decoder.push (/home/janlazo/repo/git/node-client/node_modules/msgpack-lite/lib/decode-stream.js:24:12)
```

### DIFF
--- a/__tests__/integration.test.ts
+++ b/__tests__/integration.test.ts
@@ -1,9 +1,13 @@
 /* eslint-env jest */
+import axios from 'axios';
 import * as cp from 'child_process';
 import * as fs from 'fs';
 import * as path from 'path';
+import * as util from 'util';
 
 import { attach } from '../src/index';
+
+const sleep = util.promisify(setTimeout);
 
 // eslint-disable-next-line
 jasmine.DEFAULT_TIMEOUT_INTERVAL = 5000;
@@ -35,7 +39,11 @@ describe('Node host', () => {
     fs.writeFileSync(nvimrc, `set rtp+=${integrationDir}`);
     cp.spawnSync('nvim', args);
 
-    proc = cp.spawn('nvim', ['-u', nvimrc, '-i', 'NONE', '-N', '--headless', '--embed'], {});
+    proc = cp.spawn(
+      'nvim',
+      ['-u', nvimrc, '-i', 'NONE', '-N', '--headless', '--embed'],
+      {}
+    );
     nvim = await attach({ proc });
   });
 
@@ -84,5 +92,20 @@ describe('Node host', () => {
 
   it.skip('can call a function from plugin with args', async () => {
     await nvim.command('e! nvimrc');
+  });
+
+  it('spawns a child host if $NVIM_NODE_HOST_DEBUG is set', async () => {
+    const childHost = cp.spawn(
+      process.execPath,
+      [path.join(__dirname, '..', 'bin', 'cli.js')],
+      { env: { NVIM_NODE_HOST_DEBUG: 1 }, stdio: 'ignore' }
+    );
+    await sleep(1000);
+    const debugData = await axios
+      .get('http://127.0.0.1:9229/json/list')
+      .then(res => Promise.resolve(res.data), err => Promise.reject(err));
+    childHost.kill();
+    expect(Array.isArray(debugData) && debugData.length).toBeTruthy();
+    expect(debugData[0].webSocketDebuggerUrl).toMatch('ws://127.0.0.1:9229');
   });
 });
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
     "@types/traverse": "^0.6.29",
     "@types/which": "^1.0.28",
     "@types/winston": "^2.3.3",
+    "axios": "^0.18.0",
     "babel-core": "^6.25.0",
     "eslint": "^4.19.1",
     "eslint-config-airbnb-base": "^12.1.0",
@@ -103,6 +104,7 @@
     "transform": {
       "^.+\\.tsx?$": "ts-jest"
     },
+    "testEnvironment": "node",
     "testRegex": "(/__tests__/.*|(\\.|/)(test|spec))\\.ts$",
     "coverageDirectory": "./coverage/",
     "testURL": "http://localhost"
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
   "dependencies": {
     "lodash": "^4.17.4",
     "msgpack-lite": "^0.1.26",
+    "semver": "^5.6.0",
     "traverse": "^0.6.6",
     "winston": "^2.3.1"
   },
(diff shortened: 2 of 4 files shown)


billyvg:
Yeah that seems odd, I wonder if the tests are flakey?

janlazo:
I ported Vim patches for `changetick` (8.0.0735, 8.0.1633) and `autoload/` (8.0.1377, 8.0.1378)/ in nvim. I can adjust the tests if nvim and the client are behaving as expected.

justinmk:
```
  Nvim Promise API › can deal with custom types
    expect(received).toEqual(expected)
    
    Expected value to equal:
      [""]
    Received:
      []
...
    > 106 |     expect(lines).toEqual(['']);
```

Since https://github.com/neovim/neovim/pull/8660 , `nvim_buf_get_lines` returns empty list for an _unloaded_ buffer.

Regarding these 3 failures:

```
    nvim_command: Vim:E492: Not an editor command: JSHostTestCmd
...
    nvim_call_function: Vim:E117: Unknown function: Func
...
    nvim_call_function: Vim:E117: Unknown function: Func
```

That looks like a possible result of https://github.com/neovim/neovim/pull/8371. Before https://github.com/neovim/neovim/pull/8371, some API errors were silent (didn't return to the client).

janlazo:
Build passes now so how to test `--inspect-brk`?

janlazo:
> Can you include your explanation from #46 in the commit message?

Sure.

I'll port the version checks from https://github.com/neovim/neovim/blob/master/runtime/autoload/provider/node.vim from `--inspect-brk` before writing the tests.

package.json
```diff
@@ -39,6 +39,7 @@
   "dependencies": {
     "lodash": "^4.17.4",
     "msgpack-lite": "^0.1.26",
+    "semver": "^5.6.0",
```
janlazo: I don't know why this is required. It comes with npm so the precommit check shouldn't complain. Maybe we can use node and the semver npm module to deal with semver versions in nvim's `autoload/`.

__tests__/integration.test.ts
```diff
@@ -85,4 +90,24 @@ describe('Node host', () => {
...
+    const childHost = cp.spawn(
+      process.execPath,
+      [path.join(__dirname, '..', 'bin', 'cli.js')],
+      { env: { NVIM_NODE_HOST_DEBUG: 1 } }
+    );
+    return axios.get('http://127.0.0.1:9229/json/list').then(
+      res => {
+        childHost.kill();
+        const debuggerData = res.data[0];
+        expect(debuggerData.webSocketDebuggerUrl).toMatch(
```
janlazo: I don't know if the result of this `expect` affects the test result. I can guarantee the behaviour by removing `async` because `await` is useless for http requests.

package.json
billyvg: I think it's good to have an explicit dependency for it just in case a user's environment isn't what you expect.

__tests__/integration.test.ts
```diff
@@ -85,4 +90,24 @@ describe('Node host', () => {
...
+          'ws://127.0.0.1:9229'
+        );
+      },
+      () => {
+        childHost.kill();
```
billyvg: would this be considered a test failure? should this throw an exception?

janlazo: I assume that, if the test returns a rejected promise, then Jasmine knows that the test failed. This line must run whether the promise is fulfilled or rejected so I wanted to use `Promise.prototype.finally` for this but it is not supported in Node 8 unless I use promise polyfill (or Babel?).

janlazo: If I pass the error to the reject callback, I get `ECONREFUSED`. The resolve callback doesn't run because it gets killed too early. Need sleep or timeout.

package.json
```diff
@@ -104,6 +104,7 @@
     "transform": {
       "^.+\\.tsx?$": "ts-jest"
     },
+    "testEnvironment": "node",
```
janlazo: axios was using `XMLhttpRequest` because of the `jsdom` environment so this should force axios to use node's `http`.

__tests__/integration.test.ts
```diff
@@ -95,19 +98,18 @@ describe('Node host', () => {
     const childHost = cp.spawn(
       process.execPath,
       [path.join(__dirname, '..', 'bin', 'cli.js')],
-      { env: { NVIM_NODE_HOST_DEBUG: 1 } }
+      { env: { NVIM_NODE_HOST_DEBUG: 1 }, stdio: 'ignore' }
```
janlazo: Need to close/ignore all fds to prevent child host from waiting indefinitely. `childHost.kill` is not enough.

__tests__/integration.test.ts
```diff
@@ -95,19 +98,18 @@ describe('Node host', () => {
...
-        );
-      },
-      () => {
-        childHost.kill();
+    await sleep(1000);
+    const debugData = await axios.get('http://127.0.0.1:9229/json/list').then(
+      res =>
+        Array.isArray(res.data) && res.data.length ? res.data[0] : undefined,
+      err => {
+        throw err;
```
janlazo: Not sure how `async/await` works here so I threw error from axios just to be sure. Reject callback should return undefined here so that the following `expect` assertion will catch it if throw doesn't work.

PR #96, opened 2018-10-06
off by one error

yanick:
`replace` removes one line too many

### DIFF
--- a/src/api/Buffer.test.ts
+++ b/src/api/Buffer.test.ts
@@ -277,8 +277,8 @@ describe('Buffer API', () => {
     );
 
     it('replaces the right lines', withBuffer([], async buffer => {
-        buffer.append([ '0', '1', '2', '3', '4', '5', '6', '7', '8', '9' ]);
-        buffer.replace([ 'a', 'b', 'c' ], 2)
+        await buffer.append([ '0', '1', '2', '3', '4', '5', '6', '7', '8', '9' ]);
+        await buffer.replace([ 'a', 'b', 'c' ], 2)
 
         expect(await buffer.lines).toEqual([
             '0', '1', 'a', 'b', 'c', '5', '6', '7', '8', '9' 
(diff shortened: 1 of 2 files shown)


justinmk:
Thank you! Any chance you can write a test? Looks like [Buffer.test.ts](https://github.com/neovim/node-client/blob/master/src/api/Buffer.test.ts) would be the place for it.

yanick:
:tada: Test added.

justinmk:
Actually, this change fails the _existing_ tests, though your change makes sense to me. This `replace()` function is an enhancement around Nvim's API, so I don't know what the intentions were. @billyvg  what do you think about this change?

yanick:
oops, I thought the testcase used a different nvim buffer for each test, but it seems it's not the case and I polluted the other tests. I'll try to clean that up later today.

As for the fix, I'm pretty sure the end index is off by one. Without my fix, you can see with my added test that the final buffer with end up being one line shorter every time `replace()` is called.

yanick:
Hmmm... I might be wrong, but I think the tests have been set in a rather fragile way. If we look at https://github.com/neovim/node-client/blob/master/src/api/Buffer.test.ts#L123, that test needs the buffer to be in the state left by the previous test. The problem with that is that (a) order of tests is not guaranteed to be the same as the declaration order in the file, and (b) if the test is run with `jest --watch`, then the test will likely never succeed after the first fail (as we keep adding more and more lines).

deltaskelta:
can I verify that I am not misusing the API and that since this is not in a release yet, replace is still deleting a line?

deltaskelta:
right but this line is still in master which means it is still broken because it deletes an extra line... https://github.com/neovim/node-client/blob/master/src/api/Buffer.ts#L126

yanick:
Yeah, it would make sense that the fix in this PR is not yet in master as this PR has not been merged to master yet.

justinmk:
Master was merged-into this PR , the tests are failing. Would like to merge this, but why are the tests failing?

PR #97, opened 2018-10-06
add documentation generation via typedoc

yanick:
This project is awesome, but to know the different classes / functions available, one has to dive right into the code.

So I've added a first stab at automated documentation generation via `typedoc`. The first commit of this PR is the addition of `typedoc` itself, and the second one has its output. 

if you peeps like this, we could take things one step further and make that documentation available via the github.io page of this project.


justinmk:
This is great. 

A minor concern is the size of the commit. Could that be reduced by publishing to readthedocs instead? We did that for the Nvim python client here: https://github.com/neovim/python-client/pull/302 . That's also nice because it regenerates the docs automatically.

yanick:
I'll see what was done for the python client and try to reproduce it here. 

And yeah, I'm super-familiar with the Typescript ecosystem (yet), so I picked `typedoc` because it seems to be decent, and not too painful to use. 

Something else that can be done is to only commit the produced docs in the `pages` (if I remember correct?) branch used for the github page. That way at least the generated documentation is not part of `master` and the working branches.

But first, let's see what readthedocs has to offer. I'll be back with information. :-)

yanick:
Working notes:

JS support for sphinx -- https://github.com/erikrose/sphinx-js

Also, slightly tangential: there is https://www.npmjs.com/package/typedoc-dash-theme that would generate a docset for Zeal / Dash. Right now it seems to be b0rked, but I might try to see if I can fix it.

billyvg:
Having this on the github pages branch (I don't remember what it is off the top of my head), would be great - how would we be able to automate this?

yanick:
Okay, I come back with information:

sphinx-js can generate docs for readthedocs for JS projects, and they have a beta typescript mode (that uses `typedoc` under the hood). It, however, seems to have issues with the code on this project. My proto-branch where I use it is at https://github.com/yanick/node-client/tree/doc-sphinx

Since readthedocs is a little more Python-oriented, I also looked at Gitbook. The result: https://yanick.gitbook.io/workspace/v/gitbook/_api_buffer_/_api_buffer_.buffer  (which feeds off https://github.com/yanick/node-client/tree/gitbook)

The Gitbook version of the docs takes ~560K, which is slimmer than the standard typedoc template (which is ~ 2.7M)

If you peeps like the Gitbook version, we can keep it in the work branches themselves (updating the `npm build` to incorporate its generation will then be trivial), or we can keep it to its own branch and re-generate it upon release. In that case I'll have to see what is the most painless way to have a release push the changes to that other branch upon release.

justinmk:
Looks great: https://neovim.io/node-client/

justinmk:
Eh, all of the links in the right-side column are 404'ing. 

Example: https://neovim.io/node-client/modules/_api_base_.html

billyvg:
Fixed, now to make travis autodeploy it 🙃

PR #99, opened 2018-10-24
Update tests so that CI builds pass

janlazo:
Resolve errors mentioned in https://github.com/neovim/node-client/pull/95#issuecomment-427086896

### DIFF
--- a/src/api/Buffer.test.ts
+++ b/src/api/Buffer.test.ts
@@ -28,23 +28,23 @@ describe('Buffer API', () => {
   let proc;
   let nvim: NeovimClient;
 
-  // utility to allow each test to be run in its 
+  // utility to allow each test to be run in its
   // own buffer
-  function withBuffer(lines,test) { 
+  function withBuffer(lines, test) {
     return async () => {
-        await nvim.command('new');
+      await nvim.command('new');
 
-        const buffer = await nvim.buffer;
+      const buffer = await nvim.buffer;
 
-        if( lines ) {
-            await buffer;
-            await buffer.replace(lines,0);
-        }
+      if (lines) {
+        await buffer;
+        await buffer.replace(lines, 0);
+      }
 
-        await test(buffer);
+      await test(buffer);
 
-        await nvim.command('bd! ' + buffer.id);
-    }
+      await nvim.command(`bd! ${buffer.id}`);
+    };
   }
 
   beforeAll(async done => {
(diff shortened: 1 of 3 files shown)


janlazo:
https://github.com/neovim/node-client/pull/98 abused the async arrows. Create a normal named function, class, or a plain object with function references if the object members must be modified.

src/api/Buffer.test.ts
```diff
@@ -69,159 +69,188 @@ describe('Buffer API', () => {
...
+      withBuffer([], async function buffer() {
+        buffer._name = 'foo.js';
+        expect(await buffer._name).toMatch('foo.js');
+        buffer._name = 'test2.js';
+        expect(await buffer._name).toMatch('test2.js');
+      })
+    );
+
+    it(
+      'can replace first line of buffer with a string',
```
justinmk: why all of these newlines?

janlazo: Prettifier via precommit hook.

janlazo:
> Not sure what this means, since CI already runs. What was disabled?

Nothing was disabled. What I meant was that CI builds should cover all of the tasks done by the precommit hook on top of the lints, tests done by the CI. If possible, CI can catch "incomplete" commits if there is post-processing of the changes that wasn't included into the commit. I amended a commit to add the missing hunk (caused by the pretiffier) after passing the precommit check.

Not sure if https://github.com/typicode/husky is doing the precommit checks. There are many devDependencies in the package.json and I don't use those npm packages on my projects except eslint.

src/api/Buffer.test.ts
```diff
@@ -69,159 +69,188 @@ describe('Buffer API', () => {
...
+
+        expect(await buffer.lines).toEqual([
+          'test',
+          'bar',
+          'foo',
+          'test',
+          'test',
+        ]);
+      })
+    );
```
janlazo: Added missing parenthesis from https://github.com/neovim/node-client/pull/98/

src/api/Buffer.test.ts
```diff
@@ -235,83 +264,87 @@ describe('Buffer API', () => {
...
-    ));
-
-    it('can replace nvim.buffer starting at line 1', withBuffer(
-        ['test','foo'],
-        async buffer => {
-            await buffer.replace(['bar', 'bar', 'bar'], 1);
-            expect(await buffer.lines).toEqual(['test', 'bar', 'bar', 'bar']);
-        }
-    ));
-    async () => {
```
janlazo: This async callback looks useless to me so I removed it.

src/api/Buffer.test.ts
```diff
@@ -235,83 +264,87 @@ describe('Buffer API', () => {
...
+    it(
+      'can clear the buffer',
+      withBuffer(['foo'], async buffer => {
+        await buffer.remove(0, -1);
+        // One empty line
+        expect(await buffer.length).toEqual(1);
+        expect(await buffer.lines).toEqual(['']);
+      })
+    );
+  });
```
janlazo: Added missing closing parenthesis and bracket to enclose `describe('Buffer API', ...` from https://github.com/neovim/node-client/pull/98/

src/api/Buffer.test.ts
```diff
@@ -69,159 +69,188 @@ describe('Buffer API', () => {
...
-                'bar',
-                'foo',
-                'test',
-                'test',
-            ]);
-    });
+    it(
+      'sets current buffer name to "foo.js"',
+      withBuffer([], async function buffer() {
+        buffer._name = 'foo.js';
```
janlazo: Changed callback to named `buffer` function because buffer in `withBuffer` is self-contained. Didn't try `nvim.buffer`. `name` is a reserved word for functions so I replaced it with `_name`. This change applies to other tests in this file.

src/api/Buffer.test.ts
```diff
@@ -69,159 +69,188 @@ describe('Buffer API', () => {
...
-        }
-    ));
+    it(
+      'gets the current buffer name using api chaining',
+      withBuffer([], async function buffer() {
+        await (buffer._name = 'test3.js');
+        expect(await buffer._name).toMatch('test3.js');
+
+        await (() => {
+          expect(buffer._name).toMatch('test3.js');
```
janlazo: `await` expects an expression, including callbacks. If `nvim.buffer.name` is a promise, then it's okay to use `.then` but I changed `buffer` to a named function. `buffer._name` is now a string and strings don't support Promise API with `.then`.

Issue #41, opened 2017-11-14
update README


watsoncj:
Glad to see node plugin support [landed](https://github.com/neovim/neovim/pull/7458) in Neovim 0.2.2! 🎆 

This project README still references installing node-host.

What are the new requirements for building a Neovim plugin in node? Is there a way to confirm that the engine is available?

watsoncj:
This is useful for verifying that the provider is available:

`:call health#provider#check()`

```
## Clipboard (optional)
  - OK: Clipboard tool found: pbcopy

## Python 2 provider (optional)
  - INFO: `g:python_host_prog` is not set.  Searching for python2 in the environment.
  - INFO: Executable: /usr/local/bin/python2
  - INFO: Python2 version: 2.7.14
  - INFO: python2-neovim version: 0.1.13 (outdated; from /usr/local/lib/python2.7/site-packages/neovim)
  - WARNING: Latest python2-neovim is NOT installed: 0.2.0

## Python 3 provider (optional)
  - INFO: Disabled. g:loaded_python3_provider=1

## Ruby provider (optional)
  - INFO: Ruby: ruby 2.4.2p198 (2017-09-14 revision 59899) [x86_64-darwin16]
  - WARNING: Missing "neovim" gem.
    - ADVICE:
      - Run in shell: gem install neovim
      - Is the gem bin directory in $PATH? Check `gem environment`.
      - If you are using rvm/rbenv/chruby, try "rehashing".

## Node provider (optional)
  - INFO: Node: v8.9.1
  - INFO: Host: /usr/local/bin/neovim-node-host
  - OK: Latest "neovim" npm is installed: 3.5.0
```

watsoncj:
I've gathered the following requirements for using the node provider:

* `npm install -g yarn` (not sure why yarn is required by the health check, if it really is a requirement the check should say so)
* `npm install -g neovim`
* Sample code must be built with Typescript
* `:UpdateRemotePlugins` must be run after adding a new node plugin, vim-plug does not do this but it can be done like this:
```
Plug 'sample-node-plugin', {'do':':UpdateRemotePlugins'}
```

billyvg:
* yarn should not be a requirement, I will get that fixed asap.
* I will add a boilerplate for an example plugin so people can clone and get started quickly (also maybe change sample code to be node6+ compatible?)

kutsan:
Example plugin would be nice, looking forward to.

simlrh:
[...]

PR #105, opened 2019-01-23
Docs update readme

mhartington:
This addresses feedback from #41

- Moves all plugin examples to the examples dir
- Creates a minimal plugin examples with Commands, Functions, and AutoCmds


justinmk:
Thank you @mhartington . Is there one example that can live in the README.md, which can be copy-pasted ?

PR #114, opened 2019-03-06
feat(api): Add error logging to Base.request and refactor integration tests

billyvg:
Refactor integration tests to sep package to fix circular deps for lerna.


src/api/Base.ts
```diff
@@ -83,7 +83,10 @@ export class BaseApi extends EventEmitter {
     // Not possible for ExtType classes since they are only created after transport is ready
     await this._isReady;
     this.logger.debug(`request -> neovim.api.${name}`);
-    return this[DO_REQUEST](name, args);
+    const promise = this[DO_REQUEST](name, args).catch(err => {
+      this.logger.error(`Error making request to ${name}`, err);
```
rhysd: I think we should rethrow an error here.

```javascript
this.logger.error(`Error making request to ${name}`, err);
throw err;
```

Otherwise user cannot know an error happened.

packages/neovim/src/api/Base.ts
```diff
@@ -79,20 +79,35 @@ export class BaseApi extends EventEmitter {
...
+    return this[DO_REQUEST](name, args).catch(err => {
+      const newError = new Error(err.message);
+      newError.stack = stack;
+      this.logger.error(`Error making request to ${name}`, newError);
+      throw newError;
+    });
+  }
+
+  request(name: string, args: any[] = []): Promise<any> {
+    const error = new Error(`Error making request to ${name}`);
```
billyvg: Split this method up so we can have a more helpful stacktrace.

PR #138, opened 2020-01-07
feat(attach): allow custom logger for attach neovim proc

kwonoj:
relates to https://github.com/asvetliakov/vscode-neovim/issues/134

current logger only allows customizable behavior via env variable which works for majority of node.js entrypoint, but some cases (like vs code extension) overriding env variable is not very much feasible cases. `NeovimClient` ctor already accepts custom logger, but `attach` does not exposes it - this PR expands signature to `attach` to allow specify options for customizing points. 

Instead of expanding param to directly accepts `logger`, it accepts `options` object instead for possible future customizations.

### DIFF
--- a/packages/neovim/src/attach/attach.ts
+++ b/packages/neovim/src/attach/attach.ts
@@ -2,13 +2,16 @@ import { createConnection } from 'net';
 import * as child from 'child_process';
 
 import { NeovimClient } from '../api/client';
-import { logger } from '../utils/logger';
+import { Logger } from '../utils/logger';
 
 export interface Attach {
   reader?: NodeJS.ReadableStream;
   writer?: NodeJS.WritableStream;
   proc?: NodeJS.Process | child.ChildProcess;
   socket?: string;
+  options?: {
+    logger?: Logger;
+  };
 }
 
 export function attach({
@@ -16,6 +19,7 @@ export function attach({
   writer: _writer,
   proc,
   socket,
+  options = {},
 }: Attach) {
   let writer;
   let reader;
@@ -33,7 +37,9 @@ export function attach({
   }
 
   if (writer && reader) {
-    const neovim = new NeovimClient({ logger });
+    // eslint-disable-next-line global-require
+    const loggerInstance = options.logger || require('../utils/logger').logger; // lazy load to winston only if needed
+    const neovim = new NeovimClient({ logger: loggerInstance });
     neovim.attach({
       writer,
       reader,
--- a/packages/neovim/src/utils/logger.ts
+++ b/packages/neovim/src/utils/logger.ts
@@ -2,12 +2,12 @@ import * as winston from 'winston';
 
 const level = process.env.NVIM_NODE_LOG_LEVEL || 'debug';
 
-const logger = winston.createLogger({
+const winstonLogger = winston.createLogger({
   level,
 });
 
 if (process.env.NVIM_NODE_LOG_FILE) {
-  logger.add(
+  winstonLogger.add(
     new winston.transports.File({
       filename: process.env.NVIM_NODE_LOG_FILE,
       level,
@@ -16,11 +16,12 @@ if (process.env.NVIM_NODE_LOG_FILE) {
 }
 
 if (process.env.ALLOW_CONSOLE) {
-  logger.add(
+  winstonLogger.add(
     new winston.transports.Console({
       format: winston.format.simple(),
     })
   );
 }
 
-export { logger };
+export type Logger = Pick<winston.Logger, 'info' | 'warn' | 'error' | 'debug'>;
+export const logger: Logger = winstonLogger;
--- a/packages/neovim/src/api/Base.ts
+++ b/packages/neovim/src/api/Base.ts
@@ -1,12 +1,12 @@
 import { EventEmitter } from 'events';
 
 import { Transport } from '../utils/transport';
-import { logger as loggerModule } from '../utils/logger';
+import { Logger } from '../utils/logger';
 import { VimValue } from '../types/VimValue';
 
 export interface BaseConstructorOptions {
   transport?: Transport;
-  logger?: typeof loggerModule;
+  logger?: Logger;
   data?: Buffer;
   metadata?: any;
   client?: any;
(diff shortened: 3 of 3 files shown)


packages/neovim/src/utils/logger.ts
```diff
@@ -1,26 +1,26 @@
...
-}
-
-if (process.env.ALLOW_CONSOLE) {
-  logger.add(
-    new winston.transports.Console({
-      format: winston.format.simple(),
-    })
-  );
-}
-
```
rhysd: Why is this change necessary? I think following changes would be less impact to current code base:

- don't type `logger` as `Logger` in `logger.ts` like `const logger = winston.createLogger({ ... })`
- type `loggerInstance` like `const loggerInstance: Logger = options.logger || require('../utils/logger').logger`

kwonoj: Limiting `Logger` to typed one allows to provide any custom logger, otherwise caller to `attach` should provide instance of `winston`, which is inner implementation detail of sdk leaked to public interface.

packages/neovim/src/attach/attach.ts
```diff
@@ -2,20 +2,24 @@ import { createConnection } from 'net';
 import * as child from 'child_process';
 
 import { NeovimClient } from '../api/client';
-import { logger } from '../utils/logger';
+import { Logger } from '../utils/logger';
```
rhysd: Would you also make change in `Base.ts`

from

```ts
import { logger as loggerModule } from '../utils/logger';
```
to

```ts
import { Logger } from '../utils/logger';
```

and replace `typeof loggerModule` with `Logger`?

packages/neovim/src/utils/logger.ts
kwonoj: Reason exporting limited types in logger itself is in similar crux, all of consumer of `logger.ts` only needs explicit logger interface to write log while current object always exposed full interface to winston instead.

rhysd: I mean that type of `options.logger` is still `Logger` interface. `const loggerInstance: Logger = ...` will convert the winston logger type into Logger interface.

kwonoj: Not sure if I understood q correctly, but I think `options.logger === typeof Logger` is desired, to allow `attach` can supply any logger other than winston if needed.

rhysd:
```ts
import * as winston from 'winston';

export type Logger = Pick<winston.Logger, 'info' | 'warn' | 'error' | 'debug'>;

const level = process.env.NVIM_NODE_LOG_LEVEL || 'debug';

const logger = winston.createLogger({
  level,
});

if (process.env.NVIM_NODE_LOG_FILE) {
  logger.add(
    new winston.transports.File({
      filename: process.env.NVIM_NODE_LOG_FILE,
      level,
    })
  );
}

if (process.env.ALLOW_CONSOLE) {
  logger.add(
    new winston.transports.Console({
      format: winston.format.simple(),
    })
  );
}

// Export winston.Logger class instance
export { logger };
```

and

```ts
// Export winston.Logger class instance
export { logger };

export function attach({
  reader: _reader,
  writer: _writer,
  proc,
  socket,
  options = {},
}: Attach) {
  let writer;
  let reader;

  if (socket) {
    const client = createConnection(socket);
    writer = client;
    reader = client;
  } else if (_reader && _writer) {
    writer = _writer;
    reader = _reader;
  } else if (proc) {
    writer = proc.stdin;
    reader = proc.stdout;
  }

  if (writer && reader) {
    // eslint-disable-next-line global-require
    const logger: Logger = options.logger || require('../utils/logger').logger; // lazy load to winston only if needed
    const neovim = new NeovimClient({ logger });
    neovim.attach({
      writer,
      reader,
    });
    return neovim;
  }
  throw new Error('Invalid arguments, could not attach');
}
```

kwonoj: First, my intention is limiting expored logger instance itself also not leaking details to winston. The only place in sdk requires details of winston is where it sets up transport. Agree it changes some of code flows. 

for narrowing types, it'll work as you shared but if all logger instance shares same types explicit narrowing wouldn't be needed.

If runtime code change becomes concern, code could looks like below instead:

```
const logger: Logger = winston.createLogger({
  level,
});

if (process.env.NVIM_NODE_LOG_FILE) {
  //widen types to setup transports
  (logger as winston.Logger).add(
    new winston.transports.File({
      filename: process.env.NVIM_NODE_LOG_FILE,
      level,
    })
  );
}

if (process.env.ALLOW_CONSOLE) {
   (logger as winston.Logger).add(
    new winston.transports.Console({
      format: winston.format.simple(),
    })
  );
}

// Export Logger interface, not whole winston
export { logger };
```

rhysd: How about `export { logger: logger as Logger }`? `as` is a dangerous operation which may cause type check true negative, its usage should be minimal.

kwonoj: `export` doesn't allow syntax to type narrowing. it'll make export `logger` as different name `Logger` instead, so it'll be

```
import {Logger} from '.../logger';

Logger // typeof winston.logger
```

kwonoj: actually, `logger: logger as Logger` is not supported syntax afaik.

rhysd: ok, I understood `export { logger: logger as Logger }`. I did not know that. Alternative would be

```ts
const winstonLogger = winston.createLogger({ ... });

// snip...

export const logger: Logger = winstonLogger;
```

kwonoj: yes, that'll work. I did avoid due to it requires declaring another variable reference for same instance (`winstonLogger`, `logger`).

rhysd: ya, another variable is necessary only for typing. However, I believe it would be better because, with previous code, we can only access to constructor of `winston.Logger`. When we want to operate the actual winston logger instance in the future, there is no way to access the instance.

rhysd:
@billyvg Would you also check this PR and make a new release if you're ok?

rhysd:
@kwonoj I merged. Thank you for adding this.

PR #151, opened 2020-05-01
Use macOS 10.14 on CI

rhysd:
It seems that azure pipeline has stopped support for macOS 10.13. Now only 10.14 and 10.15 are available.

https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops#use-a-microsoft-hosted-agent

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -15,7 +15,7 @@ jobs:
 - template: .azure-templates/yarn-job.yml  # Template reference
   parameters:
     name: 'macOS'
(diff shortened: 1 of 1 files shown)


rhysd:
WIP PR for checking if CI works

rhysd:
Confirmed: https://dev.azure.com/neovim/node-client/_build/results?buildId=155&view=results

rhysd:
@billyvg @justinmk Would you review and merge this? Currently CI is broken on master branch and this PR fixes it.

PR #188, opened 2021-12-13
ci: Use GitHub Actions instead of azure

billyvg:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,82 @@
...
+        ${{ matrix.EXTRACT }} ${{ matrix.NIGHTLY }}
+        echo '${{ runner.os }}'
+
+    - name: update path (bash)
+      if: runner.os != 'Windows'
+      run: echo "$(pwd)/${{ matrix.NVIM_BIN_PATH }}" >> $GITHUB_PATH
+
+    - name: update path (windows)
+      if: runner.os == 'Windows'
+      run: echo "$(pwd)/${{ matrix.NVIM_BIN_PATH }}" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
```
rhysd, line 56:
```suggestion
    - name: update path
      shell: bash
      run: echo "$(pwd)/${{ matrix.NVIM_BIN_PATH }}" >> $GITHUB_PATH
```

I'm not sure, but I guess this works. Bash is also available on Windows worker when `shell: bash` is specified.

rhysd (changes requested):

.github/workflows/ci.yml
rhysd: ok, I confirmed this comment was wrong. at 4c46391

rhysd (approved):

PR #189, opened 2022-01-06
ref: Remove `babel-cli`

billyvg:
This removes an old dep that was only used in an example package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,9 @@
     "prettier": "^2.4.1",
     "typescript": "^4.4.3"
   },
+  "resolutions": {
+    "jsprim": "1.4.2"
+  },
   "scripts": {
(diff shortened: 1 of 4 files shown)


billyvg:
This should resolve https://github.com/neovim/node-client/security/dependabot/yarn.lock/glob-parent/open

package.json
```diff
@@ -24,6 +24,9 @@
     "prettier": "^2.4.1",
     "typescript": "^4.4.3"
   },
+  "resolutions": {
+    "jsprim": "1.4.2"
+  },
```
billyvg, line 29: Resolves https://github.com/neovim/node-client/security/dependabot/yarn.lock/json-schema/open

rhysd (approved):

rhysd:
Looks good :+1:

PR #193, opened 2022-04-10
fix(ci): Fix path to nvim.exe was changed in nightly Neovim asset

rhysd:


rhysd:
All jobs passed.

https://github.com/neovim/node-client/runs/5959404572?check_suite_focus=true

PR #195, opened 2022-05-02
fix(nvim.js): $NVIM_LISTEN_ADDRESS is deprecated

justinmk:
Use $NVIM with Nvim 0.8+.

https://github.com/neovim/neovim/pull/11009 (Nvim 0.8+) will change Nvim so that $NVIM_LISTEN_ADDRESS is _not_ passed to child processes.  After that, the correct way to detect a "host" or "parent" Nvim is to check for $NVIM env var: this means the current process is a child of Nvim.

The old $NVIM_LISTEN_ADDRESS env var had conflicting purposes as both a parameter ("the current process should _listen on_ this address") and a descriptor ("the current process is a _child of_ this address").

### DIFF
--- a/packages/neovim/scripts/nvim.js
+++ b/packages/neovim/scripts/nvim.js
@@ -12,7 +12,11 @@ module.exports = (async function () {
   let proc;
   let socket;
 
-  if (process.env.NVIM_LISTEN_ADDRESS) {
+  if (process.env.NVIM) {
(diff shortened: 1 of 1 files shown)


justinmk:
I'll wait to merge this until it can run against Nvim nightly with the upstream change.

rhysd (approved):

PR #199, opened 2022-10-12
feat(api): Implement Extmark API

### LINKED ISSUE #170: Extended marks support
Do I miss something or is the support for extended marks missing?
From the [Neovim docs](https://github.com/neovim/neovim/blob/bb7d3790bf08b5519623d261d8235bad77b5c0dd/runtime/doc/api.txt#L439) I would expect that `Buffer` gets new methods/properties, e.g. `setExtmark()`.

rhysd:
Fix #170

This PR adds Extmark API binding to `Buffer`. I confirmed it worked on my local:

```javascript
const cp = require('child_process');
const { attach } = require('./packages/neovim');

const proc = cp.spawn('nvim', ['-u', 'NONE', '-N', '--embed'], {});

(async function () {
  const nvim = await attach({ proc });

  const buf = await nvim.buffer;
  await buf.replace(['this is test'], 0);

  const ns = await nvim.createNamespace('test');
  console.log('namespace:', ns);

  const id = await buf.setExtmark(ns, 0, 4);
  console.log('Mark ID:', id);

  const mark = await buf.getExtmarkById(ns, id);
  console.log('Mark:', mark);

  const marks = await buf.getExtmarks(ns, [0, 0], [0, 8]);
  console.log('Marks:', marks);

  const deleted = await buf.deleteExtmark(ns, id);
  console.log('Deleted:', deleted);

  const mark2 = await buf.getExtmarkById(ns, id);
  console.log('Mark after delete:', mark2);

  const marks2 = await buf.getExtmarks(ns, [0, 0], [0, 8]);
  console.log('Marks after delete:', marks2);

  nvim.quit();
  proc.kill();
})().catch(console.error);
```

Output:

```
namespace: 1
Mark ID: 1
Mark: [ 0, 4 ]
Marks: [ [ 1, 0, 4 ] ]
Deleted: true
Mark after delete: []
Marks after delete: []
```

### DIFF
--- a/packages/neovim/src/api/Buffer.ts
+++ b/packages/neovim/src/api/Buffer.ts
@@ -1,3 +1,4 @@
+import snakeCaseKeys = require('snakecase-keys');
 import { BaseApi } from './Base';
 import { ExtType, Metadata } from './types';
 
@@ -26,7 +27,47 @@ export interface BufferClearNamespace {
   lineEnd?: number;
 }
 
+export interface BufferGetExtmarksOptions {
+  limit?: number;
+  details?: boolean;
+}
+
+export interface BufferGetExtmarkOptions {
+  details?: boolean;
+}
+
+export interface BufferSetExtmarkOptions {
+  id?: number;
+  endRow?: number;
+  endCol?: number;
+  hlGroup?: string;
+  hlEol?: boolean;
+  virtText?: Array<[string, string | string[]]>;
+  virtTextPos?: 'eol' | 'overlay' | 'right_align';
+  virtTextWinCol?: number;
+  virtTextHide?: boolean;
+  hlMode?: 'replace' | 'combine' | 'blend';
+  virtLines?: Array<[string, string | string[]]>;
+  virtLinesAbove?: boolean;
+  virtLinesLeftcol?: boolean;
+  ephemeral?: boolean;
+  rightGravity?: boolean;
+  endRightGravity?: boolean;
+  priority?: number;
+  strict?: boolean;
+  signText?: string;
+  signHlGroup?: string;
+  numberHlGroup?: string;
+  lineHlGroup?: string;
+  cursorHlGroup?: string;
+  conceal?: string;
+  spell?: boolean;
+  uiWatched?: boolean;
+}
+
 export type VirtualTextChunk = [string, string];
+
+export type ExtmarkPosition = [number, number] | number;
 
 export const DETACH = Symbol('detachBuffer');
 export const ATTACH = Symbol('attachBuffer');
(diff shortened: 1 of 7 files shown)


justinmk:
Is this generated? We should avoid manually writing interfaces that just map to `nvim --api-info`.

rhysd:
> Is this generated?

No, it was hand-written. I agree that generating APIs from `nvim_api_info`. This time, I prioritized Extmark API since it was requested by the issue.

I think it is a bit challenging to generate APIs. For example,

- Some parameter type is just an `Array`. Its element type is unknown
- Object property is `snake_case` in Vim script but it is `lowerCamel` in TypeScript. I'm not sure it is ok to convert cases of all properties automatically

justinmk:
> * Object property is `snake_case` in Vim script but it is `lowerCamel` in TypeScript. I'm not sure it is ok to convert cases of all properties automatically

Why don't we stick to the API names exactly? TypeScript "conventions" are not that important IMO.

rhysd:
> Why don't we stick to the API names exactly? 

Since using `snake_case` looks weird. Actually no browser APIs and no Node.js APIs are using `snake_case`. Yes, it is just a convention, but I don't think it is not important.

rhysd:
OK, I'll make a new issue for generating sources. Can we merge this PR at first? Then let me try to make some mechanism of code generation. (Though I'm not sure when I can take time to do it)

justinmk:
> Some parameter type is just an Array. Its element type is unknown

This doesn't sound like a blocker. Just type the items as `any`.

I think we should strictly stick with auto-generation if we're going to provide anything beyond `request('nvim_foo', ...)`.

justinmk:
> Can we merge this PR at first?

Will we be able to remove this later once auto-generation is in place?

rhysd:
> Will we be able to remove this later once auto-generation is in place? That might be a breaking change.

I think yes. Because the auto generated sources will break many other existing APIs. So taking care about only the Extmark API doesn't make much sense.

justinmk:
> Because the auto generated sources will break many other existing APIs.

why? we already were generating stuff in the past: https://github.com/neovim/node-client/blob/master/packages/neovim/bin/generate-typescript-interfaces.js

rhysd:
> why? we already were generating stuff in the past:

If my understanding is correct, this script is no longer used. The script was used before large changes by @billyvg.

Actually, as far as I ran the script on my local machine now, it only outputs 24 lines:

```console
$ node ./bin/generate-typescript-interfaces.js                                                                                                                                                                                                                                               [Husky.local][10/13 21:56]
interface AttachOptions {
  writer?: NodeJS.WritableStream,
  reader?: NodeJS.ReadableStream,
  proc?: NodeJS.ChildProcess,
  socket?: String,
}
export default function attach(options: AttachOptions): Neovim;

export interface Neovim extends NodeJS.EventEmitter {
  quit(): void;
  isApiReady(): Boolean;
  requestApi(): Promise<[integer, any]>;
  equals(rhs: Neovim): boolean;
}
export interface Buffer {
  equals(rhs: Buffer): boolean;
}
export interface Window {
  equals(rhs: Window): boolean;
}
export interface Tabpage {
  equals(rhs: Tabpage): boolean;
}
```

rhysd:
For example, this type has information more than the output from `nvim_api_info`. So it must be hand-written.

https://github.com/neovim/node-client/blob/635b642309e7be5391d8f546f22099950f58fb24/packages/neovim/src/api/Neovim.ts#L72-L125

billyvg:
Yeah I don't think we can 100% rely on autogeneration, but it'd be good to do so in order to keep up with neovim.

@rhysd did your code comments on the PR come from neovim or were they written by you?

re: camelcase vs snake, imo we should follow node/ts conventions which is generally camelCase.

packages/neovim/src/api/Buffer.ts
```diff
@@ -381,6 +533,111 @@ export class Buffer extends BaseApi {
...
+   *
+   * Region can be given as (row,col) tuples, or valid extmark ids (whose
+   * positions define the bounds). 0 and -1 are understood as (0,0) and (-1,-1)
+   *
+   * If `end` is less than `start`, traversal works backwards. (Useful with
+   * `limit`, to get the first marks prior to a given position.)
+   *
+   * @param {Number} namespaceId Namespace id from nvim_create_namespace()
+   * @param {ExtmarkPosition} start Start of range: a 0-indexed (row, col) or valid extmark id
+   *                                (whose position defines the bound).
```
billyvg, line 549:
```suggestion
   * @param {ExtmarkPosition} start Start of range: a 0-indexed (row, col) or valid extmark id (whose position defines the bound).
```

packages/neovim/src/api/Buffer.ts
```diff
@@ -381,6 +533,111 @@ export class Buffer extends BaseApi {
...
+   * @param {ExtmarkPosition} end End of range (inclusive): a 0-indexed (row, col) or valid
+   *                              extmark id (whose position defines the bound).
```
billyvg, line 551:
```suggestion
   * @param {ExtmarkPosition} end End of range (inclusive): a 0-indexed (row, col) or valid extmark id (whose position defines the bound).
```

packages/neovim/src/api/Buffer.ts
```diff
@@ -381,6 +533,111 @@ export class Buffer extends BaseApi {
...
+   * @param {BufferGetExtmarksOptions} options Optional parameters. Keys:
+   *                                           • limit: Maximum number of marks to return
+   *                                           • details: Whether to include the details dict
```
billyvg, line 554:
```suggestion
   * @param {BufferGetExtmarksOptions} options Optional parameters. Keys:
   *    • limit: Maximum number of marks to return
   *    • details: Whether to include the details dict
```

packages/neovim/src/api/Buffer.ts
```diff
@@ -381,6 +533,111 @@ export class Buffer extends BaseApi {
...
+    ]);
+  }
+
+  /**
+   * Gets the position (0-indexed) of an extmark.
+   *
+   * @param {Number} ns_id Namespace id from |nvim_create_namespace()|
+   * @param {Number} id Extmark id
+   * @param {BufferGetExtmarkOptions} options Optional parameters. Keys:
+   *                                   • details: Whether to include the details dict
```
billyvg, line 579:
```suggestion
   *    • details: Whether to include the details dict
```

billyvg (approved):

justinmk:
It's still not clear why this is needed. https://github.com/neovim/node-client/issues/170 isn't blocked because nodejs clients can do `request('nvim_...", ...)`. 

So why rush a hand-crafted API which we won't be able to revert later (breaking change)? Why not work on generating an API instead? What is blocking that? We generate all kinds of APIs in Nvim, is there a technical issue with doing that for nodejs client?

justinmk:
This repo has lacked maintainer activity for 6+ months, which supports the point I made above about wanting to aggressively minimize the maintenance needed for this project.

`node-client` is important (used by projects like [vscode-neovim](https://github.com/vscode-neovim/vscode-neovim)). I want to keep `node-client` stable and useful, but I also want to aggressively avoid unnecessary potential for bugs and feature requests.

PR #207, opened 2023-07-11
nvim version from env

### LINKED ISSUE #146: API: "find Nvim" function
UIs and other Nvim applications often [search for Nvim](https://github.com/asvetliakov/vscode-neovim/issues/195) in the system/environment, or else they require users to specify an exact path.

- similar usecase is [firenvim's get_progpath()](https://github.com/glacambre/firenvim/blob/ca11ab668792920cac8b4195fd36b19cc32af449/autoload/firenvim.vim#L247) though that obviously requires Nvim already.

`node-client` could save a lot of hassle by providing a function like:

    get_nvim_from_env(minVersion: string)

with this behavior:

1. finds `nvim` in the current environment at a given minimum `version`
2. If nvim not found at given version, return an error that mentions the version that was found
3. return a usable path (like aforementioned firenvim `get_progpath` logic) 
   - Actually this could be in `nvim` itself. Maybe `v:progpath` should be smarter?

Frank-Mayer:
closes #146 

I'm still working on it.

The version comparison is currently just a placeholder for the correct implementation.

I would appreciate any feedback to know if I'm on the right path here. @justinmk

### DIFF
--- a/packages/neovim/src/utils/getNvimFromEnv.test.ts
+++ b/packages/neovim/src/utils/getNvimFromEnv.test.ts
@@ -1,7 +1,11 @@
 /* eslint-env jest */
 // eslint-disable-next-line import/no-extraneous-dependencies
 import which from 'which';
-import { getNvimFromEnv, compareVersions } from './getNvimFromEnv';
+import {
+  getNvimFromEnv,
+  compareVersions,
+  parseVersion,
+} from './getNvimFromEnv';
 
 try {
   which.sync('nvim');
@@ -15,6 +19,15 @@ try {
 }
 
 describe('get_nvim_from_env', () => {
+  it('Parse version', () => {
+    const a = parseVersion('0.5.0-dev+1357-g192f89ea1');
+    expect(a).toEqual([0, 5, 0, 'dev+1357-g192f89ea1']);
+    const b = parseVersion('0.5.0-dev+1357-g192f89ea1-Homebrew');
+    expect(b).toEqual([0, 5, 0, 'dev+1357-g192f89ea1-Homebrew']);
+    const c = parseVersion('0.9.1');
+    expect(c).toEqual([0, 9, 1, 'zzz']);
+  });
+
   it('Compare versions', () => {
     expect(compareVersions('0.3.0', '0.3.0')).toBe(0);
     expect(compareVersions('0.3.0', '0.3.1')).toBe(-1);
--- a/packages/neovim/src/utils/getNvimFromEnv.test.ts
+++ b/packages/neovim/src/utils/getNvimFromEnv.test.ts
@@ -1,0 +1,29 @@
+/* eslint-env jest */
+// eslint-disable-next-line import/no-extraneous-dependencies
+import * as which from 'which';
+import { getNvimFromEnv } from './getNvimFromEnv';
+
+try {
+  which.sync('nvim');
+} catch (e) {
+  // eslint-disable-next-line no-console
+  console.error(
+    'A Neovim installation is required to run the tests',
+    '(see https://github.com/neovim/neovim/wiki/Installing)'
+  );
+  process.exit(1);
+}
+
+describe('get_nvim_from_env', () => {
+  it('Gets matching nvim from specified min version', async () => {
+    const nvim = getNvimFromEnv('0.3.0');
+    expect(nvim).toBeTruthy();
+    expect(nvim).toEqual({
+      nvimVersion: expect.any(String),
+      path: expect.any(String),
+      buildType: expect.any(String),
+      luaJitVersion: expect.any(String),
+    });
+    expect(nvim!.nvimVersion).toBeTruthy();
+  });
+});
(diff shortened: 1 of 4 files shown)


Frank-Mayer:
`api/Buffer.test.ts` sometimes fails on macos 11 with node 14. However, I don't see any connection with my changes.

packages/neovim/src/utils/getNvimFromEnv.ts
```diff
@@ -0,0 +1,133 @@
...
+  const pathLength = paths.length;
+  let highestMatchingVersion: NvimVersion | null = null;
+  for (let i = 0; i !== pathLength; i = i + 1) {
+    const possibleNvimPath = join(paths[i], windows ? 'nvim.exe' : 'nvim');
+    if (existsSync(possibleNvimPath)) {
+      try {
+        accessSync(possibleNvimPath, constants.X_OK);
+        const nvimVersionFull = execSync(
+          `${possibleNvimPath} --version`
+        ).toString();
```
justinmk, line 101: `nvim --api-info` returns a msgpack object that includes a `version` dict. That avoids needing to parse `nvim --version` though I suppose that output isn't likely to change.

packages/neovim/src/utils/getNvimFromEnv.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  if (versionB.length > versionA.length) {
+    return -1;
+  }
+
+  return 0;
+}
+
+/**
+ * Get the highest matching nvim version from the environment.
```
justinmk, line 88: The "find the highest" version logic seems nice, but it might be worth dropping it to simplify the code. Instead, if the first `nvim` found doesn't meet the minimum version, the caller can simply report this to the user (mentioning the found `nvim` path and version).

fwiw, Nvim 0.9+ includes `vim.version.parse()`. Could be another way to avoid semver parsing here.

packages/neovim/src/utils/getNvimFromEnv.test.ts
justinmk, line 1: shouldn't this file live in `src/test/suite/` like the others?

packages/neovim/src/utils/getNvimFromEnv.ts
```diff
@@ -0,0 +1,133 @@
...
+        ) {
+          highestMatchingVersion = {
+            nvimVersion: nvimVersionMatch[1],
+            path: possibleNvimPath,
+            buildType: buildTypeMatch[1],
+            luaJitVersion: luaJitVersionMatch[1],
+          };
+        }
+      } catch {
+        // ignore
```
justinmk, line 128: should at least log something?

packages/neovim/src/utils/getNvimFromEnv.ts
```diff
@@ -0,0 +1,133 @@
...
+    return -1;
+  }
+
+  return 0;
+}
+
+/**
+ * Get the highest matching nvim version from the environment.
+ */
+export function getNvimFromEnv(minVersion?: string): NvimVersion | null {
```
justinmk, line 90: Would be useful to get all results, not only the best result. Let's return a list of all found items, sorted by highest version.

- callers can do `getNvimFromEnv()[0]` if they only care about the best result
- if all nvims are too old, the caller can show the user a message saying the found locations and their versions

Minor disadvantage: the caller needs to do its own "version check".

justinmk (approved):
Looks great! If you want to keep the version-comparison logic, that's not a blocker. The other comments hopefully are not much trouble.

Frank-Mayer:
Is it worth adding a dependency like [msgpackr](https://www.npmjs.com/package/msgpackr) to use the `--api-info` option instead of `--version`?

We have no control over the code in a dependency like this. This would bypass our linter rules.

I could write a more lightweight version of this that only does what we need here, but I don't think this is worth the effort.

justinmk:
> Is it worth adding a dependency like [msgpackr](https://www.npmjs.com/package/msgpackr) to use the `--api-info` option instead of `--version`?

Not worth adding a new dependency. But maybe (separate PR) worth replacing our old dependency (which was not particularly fast)? https://github.com/neovim/node-client/blob/c4ba03108dbf5360d17205f43f17a717de95cfa7/packages/neovim/package.json#L47

Frank-Mayer:
The currently used `msgpack` causes problems. I would add the feature using `--api-info` later.

Frank-Mayer:
Is there anything else you want me to change so this PR can be merged @justinmk ?

packages/neovim/src/utils/getNvimFromEnv.ts
```diff
@@ -0,0 +1,202 @@
...
+  return 0;
+}
+
+/**
+ * Get the highest matching nvim version from the environment.
+ */
+export function getNvimFromEnv(
+  opt: GetNvimFromEnvOptions = {}
+): Readonly<GetNvimFromEnvResult> {
+  const paths = process.env.PATH.split(delimiter);
```
justinmk, line 149: $PATH is likely not sufficient, especially for applications such as https://github.com/vscode-neovim/vscode-neovim which are started in GUI contexts and often (macOS, at least) the user's terminal $PATH is much different than whatever GUI apps are started with.

Thus we also need to search "common locations". These are the ones I know of:

```
/usr/local/bin/
/usr/bin
/opt/homebrew/bin
/home/linuxbrew/.linuxbrew/bin
$HOME/.linuxbrew/bin
$HOME/bin
```

justinmk:
Great work, thanks!

Merging, but I think the "common locations" will be needed for this to be maximally useful.

PR #215, opened 2023-07-17
docs: Update README

billyvg:
Small updates to README - copy README to `packages/neovim` so it is published to npm with it.


xiyaowong:
@justinmk Please review.

packages/neovim/README.md
```diff
@@ -0,0 +1,195 @@
...
+npm install -g neovim
+```
+
+A global package is required for neovim to be able to communicate with a plugin.
+
+## Usage
+This package exports a single `attach()` function which takes a pair of
+write/read streams and invokes a callback with a Nvim API object.
+
+### `attach`
```
justinmk, line 22: are we repeating info from the main readme here? is this for https://www.npmjs.com/package/neovim I guess?

can we link to this readme from the main readme, to avoid duplicating the info

PR #265, opened 2023-12-02
fix: fix types of AsyncWindow/Buffer/Tabpage

xiyaowong:
Wrong type but no error
image

Correct type but has an error
image


justinmk (approved):
Thanks! 

We really need to add more linting (eslint) to this project. Also looking at https://github.com/sindresorhus/eslint-plugin-unicorn

PR #307, opened 2024-02-06
refactor: getNvimFromEnv

justinmk:
- don't export `parseVersion`, `compareVersions`
- rename options
- add test coverage
- remove TypeErrors for conditions which cannot happen (thus aren't testable)

### DIFF
--- a/packages/neovim/src/utils/getNvimFromEnv.test.ts
+++ b/packages/neovim/src/utils/getNvimFromEnv.test.ts
@@ -1,34 +1,46 @@
 /* eslint-env jest */
 // eslint-disable-next-line import/no-extraneous-dependencies
 import which from 'which';
-import {
-  getNvimFromEnv,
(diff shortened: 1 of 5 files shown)


packages/neovim/src/utils/getNvimFromEnv.test.ts
```diff
@@ -3,42 +3,40 @@
 import which from 'which';
 import {
   getNvimFromEnv,
-  compareVersions,
-  parseVersion,
+  exportsForTesting,
 } from './getNvimFromEnv';
+const parseVersion = exportsForTesting.parseVersion
+const compareVersions = exportsForTesting.compareVersions
```
justinmk, line 9: Make these functions non-exported. If `nvim` is found then clients can use Nvim's own `vim.version` module, so there is no need for node-client to expose a half-baked "semver" implementation.

PR #309, opened 2024-02-07
docs: provide self-contained quickstart example

justinmk:


README.md
```diff
@@ -1,71 +1,95 @@
...
 
-const nvim_proc = cp.spawn('nvim', ['-u', 'NONE', '-N', '--embed'], {});
+```js
+import * as child_process from 'node:child_process'
+import * as assert from 'node:assert'
+import { attach, getNvimFromEnv } from 'neovim'
 
-// Attach to neovim process
+// Find `nvim` on the system and open a channel to it.
 (async function() {
```
justinmk, line 35: Now this example actually works and can immediately be saved to a file and executed.

Issue #267, opened 2023-12-03
findNvim should work without $PATH


justinmk:
# Problem

`findNvim` currently depends on `$PATH`. That makes it less useful for GUIs where the `$PATH` may be missing some common `nvim` install locations. https://github.com/neovim/node-client/pull/207/files#r1265953052

# Expected behavior

Search for `nvim` in common locations, e.g.

```
/usr/local/bin/
/usr/bin
/opt/homebrew/bin
/home/linuxbrew/.linuxbrew/bin
$HOME/.linuxbrew/bin
$HOME/bin

# lowest priority:
/snap/nvim/current/usr/bin/nvim
```

as well as any others that we possibly can, such as the install locations for https://github.com/microsoft/winget-cli and https://scoop.sh/ on Windows, if possible.

## Also...

Capture time cost for each search step and report that in the result? https://github.com/neovim/node-client/pull/207/files#r1264587700

justinmk:
This is kind of a blocker because `getNvimFromEnv` is useless on macOS without this.

PR #397, opened 2024-09-19
feat: add firstMatch option for findNvim

### LINKED ISSUE #370: findNvim - Add an option to "stop searching after a match"
I think this title needs no further explanation. :)

gjf7:
Closes #370 and #267

Introducing a new 'stopOnFirstMatch' option for the findNvim function, which helps reduce the number of system calls when the caller only need one valid match. Additionally, the search now includes common locations where Neovim might be installed.

### DIFF
--- a/packages/neovim/src/utils/findNvim.test.ts
+++ b/packages/neovim/src/utils/findNvim.test.ts
@@ -105,7 +105,7 @@ describe('findNvim', () => {
   });
 
   it('stops searching on first match when stopOnFirstMatch is True', () => {
-    const nvimRes = findNvim({ minVersion: '0.3.0', stopOnFirstMatch: true });
+    const nvimRes = findNvim({ minVersion: '0.3.0', firstMatch: true });
     expect(nvimRes).toEqual({
       matches: expect.any(Array),
       invalid: expect.any(Array),
--- a/packages/neovim/src/utils/findNvim.test.ts
+++ b/packages/neovim/src/utils/findNvim.test.ts
@@ -104,7 +104,7 @@ describe('findNvim', () => {
     });
   });
 
-  it('stops searching on first match when stopOnFirstMatch is True', () => {
+  it('stops searching on first match when firstMatch is True', () => {
     const nvimRes = findNvim({ minVersion: '0.3.0', firstMatch: true });
     expect(nvimRes).toEqual({
       matches: expect.any(Array),
--- a/packages/neovim/src/utils/findNvim.test.ts
+++ b/packages/neovim/src/utils/findNvim.test.ts
@@ -103,4 +103,20 @@ describe('findNvim', () => {
       error: undefined,
     });
   });
+
+  it('stops searching on first match when stopOnFirstMatch is True', () => {
+    const nvimRes = findNvim({ minVersion: '0.3.0', stopOnFirstMatch: true });
+    expect(nvimRes).toEqual({
+      matches: expect.any(Array),
+      invalid: expect.any(Array),
+    });
+    expect(nvimRes.matches.length).toEqual(1);
+    expect(nvimRes.matches[0]).toEqual({
+      nvimVersion: expect.any(String),
+      path: expect.any(String),
+      buildType: expect.any(String),
+      luaJitVersion: expect.any(String),
+      error: undefined,
+    });
+  });
 });
--- a/packages/neovim/src/utils/findNvim.ts
+++ b/packages/neovim/src/utils/findNvim.ts
@@ -31,6 +31,10 @@ export type FindNvimOptions = {
    *   - Example: `['0.4.4', '0.5.0', '0.4.3']`
    */
   readonly orderBy?: 'desc' | 'none';
+  /**
+   * (Optional) Stop searching after found a valid match
+   */
+  readonly stopOnFirstMatch?: boolean;
 };
 
 export type FindNvimResult = {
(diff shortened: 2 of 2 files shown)


packages/neovim/src/utils/findNvim.ts
```diff
@@ -31,6 +31,10 @@ export type FindNvimOptions = {
    *   - Example: `['0.4.4', '0.5.0', '0.4.3']`
    */
   readonly orderBy?: 'desc' | 'none';
+  /**
+   * (Optional) Stop searching after found a valid match
+   */
+  readonly stopOnFirstMatch?: boolean;
```
justinmk, line 37: Can use a shorter name, the docstring clarifies any questions.

justinmk:
```suggestion
  readonly firstMatch?: boolean;
```

justinmk (approved):
LGTM after naming nit. Thanks!

Future improvement would be to add Windows paths. And to allow the caller to pass specific paths (both directories and files) to search.

packages/neovim/src/utils/findNvim.ts
```diff
@@ -118,14 +122,31 @@ function compareVersions(a: string, b: string): number {
...
+  const home = process.env.HOME;
+  if (home) {
+    paths.push(`${home}/bin`, `${home}/.linuxbrew/bin`);
+  }
+
   const matches = new Array<NvimVersion>();
   const invalid = new Array<NvimVersion>();
-  for (let i = 0; i !== pathLength; i = i + 1) {
-    const nvimPath = join(paths[i], windows ? 'nvim.exe' : 'nvim');
+  const uniquePaths = new Set(paths);
```
justinmk, line 147: TODO: on Windows, this Set won't help much (need to normalize the slashes and perhaps drive letters).

packages/neovim/src/utils/findNvim.ts
gjf7: Fixed

packages/neovim/src/utils/findNvim.ts
```diff
@@ -117,35 +117,65 @@ function compareVersions(a: string, b: string): number {
...
+  if (windows) {
+    // Scoop common install location
+    if (USERPROFILE) {
+      paths.add(normalizePath(`${USERPROFILE}\\scoop\\shims`));
+    }
+    paths.add(normalizePath('C:\\ProgramData\\scoop\\shims'));
+
+    // Winget common install location
+    if (LOCALAPPDATA) {
+      paths.add(normalizePath(`${LOCALAPPDATA}\\Microsoft\\WindowsApps`));
```
gjf7, line 138: I'm not sure this is right default install location for winget. I tested it on my window virtual machine and it turns out neovim was installed in `C:ProgramFiles\Neovim\bin`.

I also found [this](https://github.com/microsoft/winget-cli/blob/master/doc/specs/%23182%20-%20Support%20for%20installation%20of%20portable%20standalone%20apps.md) from their docs.

packages/neovim/src/utils/findNvim.ts
gjf7: Fixed

packages/neovim/src/utils/findNvim.test.ts
```diff
@@ -103,4 +103,20 @@ describe('findNvim', () => {
       error: undefined,
     });
   });
+
+  it('stops searching on first match when stopOnFirstMatch is True', () => {
```
justinmk, line 107:
```suggestion
  it('stops searching on first match when firstMatch is True', () => {
```

packages/neovim/src/utils/findNvim.ts
```diff
@@ -113,19 +117,66 @@ function compareVersions(a: string, b: string): number {
...
+
+  PATH?.split(delimiter).forEach(p => paths.add(normalizePath(p)));
+
+  // Add common Neovim installation paths not always in the system's PATH.
+  if (windows) {
+    // Scoop common install location
+    if (USERPROFILE) {
+      paths.add(normalizePath(`${USERPROFILE}\\scoop\\shims`));
+    }
+    paths.add(normalizePath('C:\\ProgramData\\scoop\\shims'));
```
justinmk, line 134: no need to use \ slashes, even on windows
```suggestion
    paths.add(normalizePath('C:/ProgramData\/scoop/shims'));
```

packages/neovim/src/utils/findNvim.ts
```diff
@@ -113,19 +117,66 @@ function compareVersions(a: string, b: string): number {
   return 0;
 }
 
+function getPlatformPaths() {
+  const paths = new Set<string>();
+  const { PATH, USERPROFILE, LOCALAPPDATA, PROGRAMFILES, HOME } = process.env;
+
+  const normalizePath = (path: string) => normalize(path.toLowerCase());
```
justinmk, line 124:
```suggestion
  const normalizePath = (path: string) => normalize(windows ? path.toLowerCase() : path);
```

packages/neovim/src/utils/findNvim.ts
```diff
@@ -113,19 +117,66 @@ function compareVersions(a: string, b: string): number {
...
+ * @param opt.firstMatch See {@link FindNvimOptions.firstMatch}
  */
-export function findNvim(opt: FindNvimOptions = {}): Readonly<FindNvimResult> {
-  const paths = process.env.PATH?.split(delimiter) ?? [];
-  const pathLength = paths.length;
+export function findNvim({
+  minVersion,
+  orderBy,
+  firstMatch,
+}: FindNvimOptions = {}): Readonly<FindNvimResult> {
```
justinmk, line 173: would you mind keeping the `opt`

packages/neovim/src/utils/findNvim.ts
justinmk: It's worth a try, thanks for researching that! Would be good to mention that URL in the comment too.

justinmk:
The codecov complaints about the Windows codepaths, should be fixed after https://github.com/neovim/node-client/pull/398

packages/neovim/src/utils/findNvim.ts
gjf7: Sure.

packages/neovim/src/utils/findNvim.ts
```diff
@@ -137,9 +189,8 @@ export function findNvim(opt: FindNvimOptions = {}): Readonly<FindNvimResult> {
         const luaJitVersionMatch = luaJitVersionRegex.exec(nvimVersionFull);
         if (nvimVersionMatch && buildTypeMatch && luaJitVersionMatch) {
           if (
-            'minVersion' in opt &&
-            compareVersions(opt.minVersion ?? '0.0.0', nvimVersionMatch[1]) ===
-              1
+            opt.minVersion &&
+            compareVersions(opt.minVersion, nvimVersionMatch[1]) === 1
```
justinmk, line 193: the old logic supported callers that explicitly pass `minVersion = undefined`.

gjf7: Oops, my bad

justinmk (approved):

gjf7:
@justinmk Thanks for your patient review! This is my first time contributing to node-client, and I'm so excited.

justinmk:
It's great to have some help around here :)

PR #400, opened 2024-09-22
feat: add paths and dirs options for findNvim

### LINKED ISSUE #399: findNvim: caller can supply explicit paths, dirs
## Problem

Apps such as https://github.com/vscode-neovim/vscode-neovim use `findNvim` to find Nvim on the user's system, but:

- they also support user settings that allow the user to explicitly set a path.
- the app may want to search other directories besides the default ones hardcoded in `findNvim`

Since `findNvim` currently doesn't allow the caller to provide extra paths/directories, consumers such as vscode-neovim have to re-invent their own way to validate those extra paths/directories, which is redundant with what `findNvim` already does.

## Proposal

Introduce `paths: string[]` and `dirs: string[]` options to `findNvim`.

gjf7:
Closes #399.

### DIFF
--- a/packages/neovim/src/utils/findNvim.test.ts
+++ b/packages/neovim/src/utils/findNvim.test.ts
@@ -1,12 +1,26 @@
 /* eslint-env jest */
 import { join } from 'node:path';
-import * as fs from 'node:fs';
+import { mkdirSync, writeFileSync, rmSync } from 'node:fs';
 import { findNvim, exportsForTesting, FindNvimResult } from './findNvim';
 
 const parseVersion = exportsForTesting.parseVersion;
 const compareVersions = exportsForTesting.compareVersions;
 
 describe('findNvim', () => {
+  const testDir = join(process.cwd(), 'test-dir');
+  const nvimExecutablePath = join(
+    testDir,
+    process.platform === 'win32' ? 'nvim.exe' : 'nvim'
+  );
+
+  beforeAll(() => {
+    mkdirSync(testDir, { recursive: true });
+    writeFileSync(nvimExecutablePath, 'fake-nvim-executable');
+  });
+
+  afterAll(() => {
+    rmSync(testDir, { recursive: true, force: true });
+  });
   it('parseVersion()', () => {
     expect(parseVersion('0.5.0-dev+1357-g192f89ea1')).toEqual([
       0,
@@ -123,26 +137,26 @@ describe('findNvim', () => {
   });
 
   it('searches in additional custom paths', () => {
-    const customPaths = ['/custom/path/to/nvim', '/another/custom/path'];
-
-    jest
-      .spyOn(fs, 'existsSync')
-      .mockImplementation(path => customPaths.includes(path as string));
-
+    const customPaths = [
+      `${process.cwd()}/package.json`,
+      '/custom/path/to/nvim',
+      '/another/custom/path',
+    ];
     const nvimRes = findNvim({ paths: customPaths });
-    assertOneOrMore(nvimRes);
+
+    expect(nvimRes.matches.length).toBeGreaterThanOrEqual(1);
+
+    expect(nvimRes.invalid.length).toBe(1);
+    expect(nvimRes.invalid[0].path).toEqual(customPaths[0]);
   });
 
-  it('searches in additional custom directories', () => {
-    const customDirs = ['/custom/dir', '/another/custom/dir'];
-    const nvimExecutable = process.platform === 'win32' ? 'nvim.exe' : 'nvim';
-    const customPaths = customDirs.map(dir => join(dir, nvimExecutable));
-
-    jest
-      .spyOn(fs, 'existsSync')
-      .mockImplementation(path => customPaths.includes(path as string));
-
+  it('searches in additional custom dirs', () => {
+    const customDirs = [testDir, '/non/existent/dir'];
     const nvimRes = findNvim({ dirs: customDirs });
-    assertOneOrMore(nvimRes);
+
+    expect(nvimRes.matches.length).toBeGreaterThanOrEqual(1);
+
+    expect(nvimRes.invalid.length).toBe(1);
+    expect(nvimRes.invalid.map(r => r.path)).toContain(nvimExecutablePath);
   });
 });
--- a/packages/neovim/src/utils/findNvim.test.ts
+++ b/packages/neovim/src/utils/findNvim.test.ts
@@ -5,12 +5,12 @@ import { findNvim, exportsForTesting, FindNvimResult } from './findNvim';
 
 const parseVersion = exportsForTesting.parseVersion;
 const compareVersions = exportsForTesting.compareVersions;
+const normalizePath = exportsForTesting.normalizePath as (p: string) => string;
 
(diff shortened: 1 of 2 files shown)


packages/neovim/src/utils/findNvim.test.ts
```diff
@@ -119,4 +121,28 @@ describe('findNvim', () => {
       error: undefined,
     });
   });
+
+  it('searches in additional custom paths', () => {
+    const customPaths = ['/custom/path/to/nvim', '/another/custom/path'];
+
+    jest
+      .spyOn(fs, 'existsSync')
+      .mockImplementation(path => customPaths.includes(path as string));
```
justinmk, line 130: I'm planning to make findNvim "async", and then this might fail. That's an example of why spies/mocks are brittle.

To test actual behavior, we could specify real paths in the `customPath`, paths that we know exist but aren't valid: e.g. `['README.md', ...]` and then assert the results, which should contain an `invalid` object. 

As a bonus that also allows us to better test the behavior of `invalid`.

packages/neovim/src/utils/findNvim.ts
```diff
@@ -35,6 +35,23 @@ export type FindNvimOptions = {
    * (Optional) Stop searching after found a valid match
    */
   readonly firstMatch?: boolean;
+  /**
+   * (Optional) Additional specific file paths to check for Nvim executables.
+   * These paths will be checked before searching in directories.
```
justinmk, line 40:
```suggestion
   * These paths will be checked before searching `dirs`.
```

packages/neovim/src/utils/findNvim.ts
```diff
@@ -35,6 +35,23 @@ export type FindNvimOptions = {
...
+   * Useful for allowing users to specify exact Nvim executable locations.
+   *
+   * Example: ['/usr/local/bin/nvim', '/opt/homebrew/bin/nvim']
+   */
+  readonly paths?: string[];
+  /**
+   * (Optional) Additional directories to search for Nvim executables.
+   * These directories will be searched after checking specific paths
+   * but before searching in default system locations.
```
justinmk, line 49:
```suggestion
   * These directories will be searched after checking `paths`
   * but before searching `$PATH` and other default locations.
```

packages/neovim/src/utils/findNvim.ts
```diff
@@ -117,7 +134,7 @@ function compareVersions(a: string, b: string): number {
   return 0;
 }
 
-function getPlatformPaths() {
+function getPlatformSearchDirectories(): Set<string> {
```
justinmk, line 137:
```suggestion
function getPlatformSearchDirs(): Set<string> {
```

packages/neovim/src/utils/findNvim.ts
```diff
@@ -170,14 +188,22 @@ function getPlatformPaths() {
...
+ * @param opt.dirs See {@link FindNvimOptions.dirs}
  */
 export function findNvim(opt: FindNvimOptions = {}): Readonly<FindNvimResult> {
-  const paths = getPlatformPaths();
+  const platformDirs = getPlatformSearchDirectories();
+  const nvimExecutable = windows ? 'nvim.exe' : 'nvim';
 
+  const allPaths = new Set<string>([
+    ...(opt.paths || []),
+    ...[...(opt.dirs || []), ...platformDirs].map(dir =>
```
justinmk, line 200: should we normalize opts.paths/dirs to avoid duplicates in the Set?

gjf7: Good point, I think we should.

packages/neovim/src/utils/findNvim.ts
gjf7: Fixed

packages/neovim/src/utils/findNvim.ts
gjf7: Fixed

packages/neovim/src/utils/findNvim.ts
gjf7: Fixed

gjf7:
Finally makes ci happy.

packages/neovim/src/utils/findNvim.ts
```diff
@@ -170,14 +189,22 @@ function getPlatformPaths() {
...
+ * @param opt.dirs See {@link FindNvimOptions.dirs}
  */
 export function findNvim(opt: FindNvimOptions = {}): Readonly<FindNvimResult> {
-  const paths = getPlatformPaths();
+  const platformDirs = getPlatformSearchDirs();
+  const nvimExecutable = windows ? 'nvim.exe' : 'nvim';
+
+  const allPaths = new Set<string>([
+    ...(opt.paths || []).map(normalizePath),
+    ...(opt.dirs || []).map(dir => normalizePath(join(dir, nvimExecutable))),
```
justinmk, line 201: || is different than ??
```suggestion
    ...(opt.paths ?? []).map(normalizePath),
    ...(opt.dirs ?? []).map(dir => normalizePath(join(dir, nvimExecutable))),
```

packages/neovim/src/utils/findNvim.test.ts
```diff
@@ -119,4 +135,32 @@ describe('findNvim', () => {
...
+    const customPaths = [
+      join(process.cwd(), 'package.json'),
+      '/custom/path/to/nvim',
+      '/another/custom/path',
+    ];
+    const nvimRes = findNvim({ paths: customPaths });
+
+    expect(nvimRes.matches.length).toBeGreaterThanOrEqual(1);
+
+    expect(nvimRes.invalid.length).toBe(1);
```
justinmk, line 149: i would except this to be "3" because we passed 3 invalid paths. why is it 1?

packages/neovim/src/utils/findNvim.ts
```diff
@@ -170,14 +189,22 @@ function getPlatformPaths() {
...
+    ...(opt.dirs ?? []).map(dir => normalizePath(join(dir, nvimExecutable))),
+    ...[...platformDirs].map(dir => join(dir, nvimExecutable)),
+  ]);
 
   const matches = new Array<NvimVersion>();
   const invalid = new Array<NvimVersion>();
-  for (const path of paths) {
-    const nvimPath = join(path, windows ? 'nvim.exe' : 'nvim');
+  for (const nvimPath of allPaths) {
     if (existsSync(nvimPath)) {
```
justinmk, line 208: oh, this was for the case where we were searching directories. in that case, we still want to ignore "not found" because we are manually building the path by appending `nvim`.

but for the case where `paths` is explicitly passed, I think it makes sense to add any/all of those paths to `invalid` if they are not found.

justinmk, line 208: maybe this works;

```suggestion
    if (existsSync(nvimPath) || opt.paths?.includes(nvimPath)) {
```

gjf7: For "not found", what error should we put in error field in invalid?

gjf7: Yeah, this works

justinmk:
Thanks again!

Issue #352, opened 2024-04-17
build: enable full typescript checking of test files


justinmk:
# Problem

currently, test files are not checked by `npm run build`, we exclude them in tsconfig: https://github.com/neovim/node-client/blob/d3bc8d36b50ccc188e083b33ff537819e661dd88/packages/neovim/tsconfig.json#L7

# Expected behavior

- test files should be checked
- fix all the errors

PR #410, opened 2024-10-08
build(tests): type check Buffer.test.ts

gjf7:
ref #352

Our linter and TypeScript config disagreed on test files. Linter included them, config didn't.

https://github.com/neovim/node-client/blob/65ddb5e7c8fa6c1bd4a6cd6564ad0cdf83b7e92c/packages/neovim/tsconfig.json#L7

https://github.com/neovim/node-client/blob/65ddb5e7c8fa6c1bd4a6cd6564ad0cdf83b7e92c/package.json#L37

This PR should fix this as well.

### DIFF
--- a/packages/neovim/src/api/Buffer.test.ts
+++ b/packages/neovim/src/api/Buffer.test.ts
@@ -1,5 +1,7 @@
 /* eslint-env jest */
+import assert from 'node:assert';
 import * as testUtil from '../testUtil';
+import type { Buffer } from './Buffer';
 
 function wait(ms: number): Promise<void> {
   return new Promise(resolve => {
@@ -14,14 +16,16 @@ describe('Buffer API', () => {
 
   // utility to allow each test to be run in its
   // own buffer
-  function withBuffer(lines, test) {
+  function withBuffer(
+    lines: string[],
+    test: (buffer: Buffer) => Promise<void>
+  ) {
     return async () => {
       await nvim.command('new!');
 
       const buffer = await nvim.buffer;
 
       if (lines) {
-        await buffer;
         await buffer.replace(lines, 0);
       }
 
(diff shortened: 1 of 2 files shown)


justinmk:
Thanks for helping with this! Probably want to do a couple files at a time, not one big PR.

packages/neovim/src/api/Buffer.test.ts
```diff
@@ -216,7 +220,7 @@ describe('Buffer API', () => {
         expect(await buffer.getOption('copyindent')).toBe(false);
 
         // Restore option
-        buffer.setOption('copyindent', initial);
+        buffer.setOption('copyindent', initial!);
```
justinmk, line 223: nit: could at assert(initial) on line 217, that might be a better pattern in general.

packages/neovim/tsconfig.json
```diff
@@ -4,5 +4,5 @@
     "outDir": "./lib"
   },
   "include": ["src"],
-  "exclude": ["node_modules", "__tests__", "**/*.test.ts", "examples", "lib"]
+  "exclude": ["node_modules", "examples", "lib"]
```
justinmk, line 7: to unblock this PR so we can work incrementally, maybe leave the old exclude and add specific `include` for the fixed files, e.g.:

    "include": ["src", "**/api/Buffer.test.ts"],

justinmk (approved):

packages/neovim/src/api/Buffer.test.ts
gjf7: Use assert for type narrowing? I didn't know that. Thanks for teaching me something new!

packages/neovim/tsconfig.json
gjf7: Fixed

PR #414, opened 2024-10-11
test(attach): improve disconnect test

### LINKED ISSUE #314: unreliable test: `emits "disconnect" after quit`
```
FAIL src/attach/attach.test.ts (6.169 s)
  ● Nvim Promise API › emits "disconnect" after quit

    thrown: "Exceeded timeout of 5000 ms for a test while waiting for `done()` to be called.
    Add a timeout value to this test to increase the timeout, if this is a long-running test. See https://jestjs.io/docs/api#testname-fn-timeout."

      104 |   });
      105 |
    > 106 |   it('emits "disconnect" after quit', done => {
          |   ^
      107 |     const disconnectMock = jest.fn();
      108 |     nvim.on('disconnect', disconnectMock);
      109 |     nvim.quit();

      at it (src/attach/attach.test.ts:106:3)
      at Object.describe (src/attach/attach.test.ts:18:1)

A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --detectOpenHandles to find leaks. Active timers can also cause this, ensure that .unref() was called on them.
```

gjf7:
Closes #314.

Not sure if this is a good improvement, any suggestions? @justinmk

### DIFF
--- a/packages/neovim/src/attach/attach.test.ts
+++ b/packages/neovim/src/attach/attach.test.ts
@@ -160,6 +160,7 @@ describe('Nvim API', () => {
 
     nvim.quit();
 
+    // TODO: 'close' event sometimes does not emit. #414
     proc.on('exit', () => {
       expect(disconnectMock).toHaveBeenCalledTimes(1);
       done();
--- a/packages/neovim/src/attach/attach.test.ts
+++ b/packages/neovim/src/attach/attach.test.ts
@@ -154,27 +154,22 @@ describe('Nvim API', () => {
     expect(newLines).toEqual(['line1', 'line2']);
   });
 
-  it('emits "disconnect" after quit', async () => {
-    const disconnectMock = jest.fn();
-    nvim.on('disconnect', disconnectMock);
-    nvim.quit();
-
-    const closePromise = new Promise(resolve => {
-      proc.on('close', resolve);
+  it('emits "disconnect" after quit', done => {
+    let called = false;
+    nvim.on('disconnect', () => {
+      called = true;
     });
 
-    let timeout: NodeJS.Timeout;
+    nvim.quit();
 
-    await Promise.race([
-      closePromise.then(() => clearTimeout(timeout)),
-      new Promise((_, reject) => {
-        timeout = setTimeout(
-          () => reject(new Error('Timeout waiting for nvim to quit')),
-          5000
-        );
-      }),
-    ]);
+    proc.on('close', () => {
+      expect(called).toBe(true);
+      done();
+    });
 
-    expect(disconnectMock).toHaveBeenCalledTimes(1);
-  }, 10000);
+    // Event doesn't actually emit when we quit nvim, but when the child process is killed
+    if (proc && proc.connected) {
+      proc.disconnect();
+    }
+  });
 });
--- a/packages/neovim/src/attach/attach.test.ts
+++ b/packages/neovim/src/attach/attach.test.ts
@@ -154,19 +154,27 @@ describe('Nvim API', () => {
     expect(newLines).toEqual(['line1', 'line2']);
   });
 
-  it('emits "disconnect" after quit', done => {
+  it('emits "disconnect" after quit', async () => {
     const disconnectMock = jest.fn();
     nvim.on('disconnect', disconnectMock);
     nvim.quit();
 
-    proc.on('close', () => {
-      expect(disconnectMock.mock.calls.length).toBe(1);
-      done();
+    const closePromise = new Promise(resolve => {
+      proc.on('close', resolve);
     });
 
-    // Event doesn't actually emit when we quit nvim, but when the child process is killed
-    if (proc && proc.connected) {
-      proc.disconnect();
-    }
-  });
+    let timeout: NodeJS.Timeout;
+
+    await Promise.race([
+      closePromise.then(() => clearTimeout(timeout)),
+      new Promise((_, reject) => {
+        timeout = setTimeout(
+          () => reject(new Error('Timeout waiting for nvim to quit')),
+          5000
+        );
+      }),
+    ]);
+
+    expect(disconnectMock).toHaveBeenCalledTimes(1);
+  }, 10000);
 });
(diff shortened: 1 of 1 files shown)


packages/neovim/src/attach/attach.test.ts
```diff
@@ -154,19 +154,27 @@ describe('Nvim API', () => {
...
+      new Promise((_, reject) => {
+        timeout = setTimeout(
+          () => reject(new Error('Timeout waiting for nvim to quit')),
+          5000
+        );
+      }),
+    ]);
+
+    expect(disconnectMock).toHaveBeenCalledTimes(1);
   });
```
gjf7, line 179: Should I increase timeout of this test, like 10s ?

justinmk: No, most likely either the test harness (jest) is interfering with something (jest is known to proxy the `process` but idk if that affects `proc.on` here), or there's a race (bug) in our EventEmitter implementation, or we need to rethink the approach of this test.

packages/neovim/src/attach/attach.test.ts
```diff
@@ -154,19 +154,27 @@ describe('Nvim API', () => {
...
+      proc.on('close', resolve);
     });
 
-    // Event doesn't actually emit when we quit nvim, but when the child process is killed
-    if (proc && proc.connected) {
-      proc.disconnect();
-    }
+    let timeout: NodeJS.Timeout;
+
+    await Promise.race([
```
justinmk, line 168: Thanks for looking at this. I think the `done` mechanism works, the problem is either in our test logic or there's a race during "teardown" so e.g. the EventEmitter event is not getting triggered.

Or maybe it's related to jest's "mocking" implementation, so removing the use of `jest.fn` could help (for most of our tests we have no reason to use it, especially this one).

gjf7: If we remove the mock function, how do we make sure `disconnect` is indeed emitted?

justinmk: just define a real function: 

```diff
diff --git a/packages/neovim/src/attach/attach.test.ts b/packages/neovim/src/attach/attach.test.ts
index 0ae7cc029623..1146b13e9263 100644
--- a/packages/neovim/src/attach/attach.test.ts
+++ b/packages/neovim/src/attach/attach.test.ts
@@ -183,12 +183,14 @@ describe('Nvim API', () => {
   });
 
   it('emits "disconnect" after quit', done => {
-    const disconnectMock = jest.fn();
-    nvim.on('disconnect', disconnectMock);
+    let calls = 0;
+    nvim.on('disconnect', function() {
+      calls = calls + 1;
+    });
     nvim.quit();
 
     proc.on('close', () => {
-      expect(disconnectMock.mock.calls.length).toBe(1);
+      expect(calls).toBe(1);
       done();
     });
 ```

however now that I look at it, it's obvious that the mock isn't the issue here, though I think it's an unnecessary layer to introduce to the test.

File: packages/neovim/src/attach/attach.test.ts [hunk 3]
```diff
@@ -155,12 +155,15 @@ describe('Nvim API', () => {
...
+    nvim.on('disconnect', () => {
+      disconnectCalled = true;
+    });
+
     nvim.quit();
 
     proc.on('close', () => {
-      expect(disconnectMock.mock.calls.length).toBe(1);
+      expect(disconnectCalled).toBe(true);
       done();
```
justinmk, line 167: since the test is timing out, that means this `done` is not called, which means the 'close' event isn't emitted (or stuff is getting disposed before handling the emitted 'close' event).

packages/neovim/src/attach/attach.test.ts
```diff
@@ -155,12 +155,15 @@ describe('Nvim API', () => {
   });
 
   it('emits "disconnect" after quit', done => {
-    const disconnectMock = jest.fn();
-    nvim.on('disconnect', disconnectMock);
+    let disconnectCalled = false;
```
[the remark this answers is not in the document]
gjf7: Could we rerun ci couple more times to see the job failed again or not ?

packages/neovim/src/attach/attach.test.ts
gjf7: Yeah, I understand that, but how race could happen to cause this problem?

gjf7:
Oops, failed again, so the mock function indeed not the issue here.

packages/neovim/src/attach/attach.test.ts
gjf7: Do we still need the disconnect call on proc ?

justinmk: oh, that causes the test harness (jest) to kill Nvim too early, which then raises EPIPE because Nvim still has messages to process.

packages/neovim/src/attach/attach.test.ts
```diff
@@ -155,12 +155,15 @@ describe('Nvim API', () => {
...
-    const disconnectMock = jest.fn();
-    nvim.on('disconnect', disconnectMock);
+    let called = false;
+    nvim.on('disconnect', () => {
+      called = true;
+    });
+
     nvim.quit();
 
     proc.on('close', () => {
```
justinmk, line 165: Based on local runs, changing this to 'exit' seems more reliable. It's strange that 'close' is sometimes not emitted, but that was never the purpose of this test...
```suggestion
    proc.on('exit', () => {
```

gjf7: Make sense, we only want make sure disconnect is called.

gjf7:
We can run CI a few more times to see if it's working.

gjf7:
Looks like it happens on windows. see https://github.com/nodejs/node/issues/4500

packages/neovim/src/attach/attach.test.ts
```diff
@@ -157,10 +157,11 @@ describe('Nvim API', () => {
   it('emits "disconnect" after quit', done => {
     const disconnectMock = jest.fn();
     nvim.on('disconnect', disconnectMock);
+
     nvim.quit();
 
-    proc.on('close', () => {
-      expect(disconnectMock.mock.calls.length).toBe(1);
+    proc.on('exit', () => {
```
justinmk, line 163:
```suggestion
    // TODO: 'close' event sometimes does not emit. #414
    proc.on('exit', () => {
```

gjf7:
@justinmk Thanks for your guide!

PR #420, opened 2024-10-11
test: replace jest with mocha

### LINKED ISSUE #333: tests: replace jest with mocha or something else
jest fucking sucks. setting up a global beforeAll/afterAll is a pain in the ass. 

this article is correct: https://ak-88.medium.com/jest-sucks-c4822baf2073

gjf7:
Closes #333.

To make a transition from Jest feasible. Capabilities Jest offers that we might need to keep:

## Expect
Expect is a standalone npm package.

## Mock
The jest.fn(), jest.spyOn() is provided by jest-mock, we can just use this package directly.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -14,6 +14,7 @@ module.exports = {
   env: {
     node: true,
     es2024: true,
+    mocha: true,
   },
 
   overrides: [
@@ -70,6 +71,7 @@ module.exports = {
 
     'import/extensions': 'off',
     'import/prefer-default-export': 'off',
+    'import/no-extraneous-dependencies': ['error', {'devDependencies': true, 'optionalDependencies': false, 'peerDependencies': false}],
 
     '@typescript-eslint/no-namespace': 'error',
     // TODO: '@typescript-eslint/no-floating-promises': 'error', // Promises must catch errors or be awaited.
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -29,6 +29,7 @@ module.exports = {
         // `jest` against the compiled .js results (would require compiling
         // the test files as well)?
         'unicorn/prefer-at': 'off',
+        'import/no-extraneous-dependencies': ['error', {'devDependencies': true, 'optionalDependencies': false, 'peerDependencies': false}],
       },
     },
   ],
@@ -71,7 +72,6 @@ module.exports = {
 
     'import/extensions': 'off',
     'import/prefer-default-export': 'off',
-    'import/no-extraneous-dependencies': ['error', {'devDependencies': true, 'optionalDependencies': false, 'peerDependencies': false}],
 
     '@typescript-eslint/no-namespace': 'error',
     // TODO: '@typescript-eslint/no-floating-promises': 'error', // Promises must catch errors or be awaited.
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "dev": "npm run --stream --parallel dev --workspaces --if-present",
     "publish:neovim": "cd packages/neovim && cp ../../README.md . && cp ../../CHANGELOG.md . && npm publish",
     "test": "npm run test --workspaces --if-present",
-    "test-coverage": "npm run test --workspaces --if-present",
+    "test-coverage": "npm run test-coverage --workspaces --if-present",
     "test-staged": "npm run test-staged --workspaces --if-present -- --",
     "test-missing-apis": "npm run test-missing-apis --workspaces --if-present",
     "test-lint": "npm run lint",
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "dev": "npm run --stream --parallel dev --workspaces --if-present",
     "publish:neovim": "cd packages/neovim && cp ../../README.md . && cp ../../CHANGELOG.md . && npm publish",
     "test": "npm run test --workspaces --if-present",
-    "test-build": "npm run test-build --workspaces --if-present",
+    "test-coverage": "npm run test --workspaces --if-present",
     "test-staged": "npm run test-staged --workspaces --if-present -- --",
     "test-missing-apis": "npm run test-missing-apis --workspaces --if-present",
     "test-lint": "npm run lint",
--- a/packages/integration-tests/package.json
+++ b/packages/integration-tests/package.json
@@ -35,8 +35,8 @@
     "neovim": "file:../neovim"
   },
   "scripts": {
-    "test": "jest",
-    "test-build": "npm test --runInBand --coverage",
+    "test": "mocha --exit --require ts-node/register src/**/*.test.ts __tests__/**/*.test.ts",
+    "test-coverage": "c8 --reporter=html --reporter=json npm test",
     "build": "tsc --pretty",
     "dev": "npm run build --watch true"
   },
@@ -50,12 +50,10 @@
     "vim"
   ],
   "devDependencies": {
-    "jest": "^29.7.0",
-    "jest-haste-map": "^29.7.0",
-    "jest-resolve": "^29.6.1"
-  },
-  "jest": {
-    "testEnvironment": "node",
-    "testPathIgnorePatterns": [".d.ts", ".js"]
+    "@types/mocha": "^10.0.9",
+    "c8": "^10.1.2",
+    "expect": "^29.7.0",
+    "jest-mock": "^29.7.0",
+    "mocha": "^10.7.3"
   }
 }
--- a/packages/neovim/package.json
+++ b/packages/neovim/package.json
@@ -71,7 +71,7 @@
   "scripts": {
     "doc": "typedoc --out doc --exclude '**/*.test.ts' src",
     "prepublishOnly": "npm run build",
-    "test": "NODE_ENV=test mocha --exit --require ts-node/register --require src/testSetup.ts src/**/*.test.ts",
+    "test": "mocha",
     "test-coverage": "c8 --reporter=json --reporter=html npm test",
     "test-staged": "npm test --bail",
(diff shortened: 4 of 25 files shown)


gjf7:
@justinmk Mocha doesn't support coverage out of box, Do u think https://github.com/istanbuljs/nyc is a good choice ?

justinmk:
> @justinmk Mocha doesn't support coverage out of box, Do u think https://github.com/istanbuljs/nyc is a good choice for us?

nyc is deprecated, https://github.com/bcoe/c8 is the current best option AFAIK.

packages/neovim/package.json
```diff
@@ -58,50 +58,27 @@
...
   },
   "scripts": {
     "doc": "typedoc --out doc --exclude '**/*.test.ts' src",
     "prepublishOnly": "npm run build",
-    "test": "jest",
-    "test-build": "npm test --runInBand --coverage",
-    "test-staged": "npm test --bail --no-cache --findRelatedTests",
+    "test": "mocha --exit --parallel --require ts-node/register --require src/testSetup.ts src/**/*.test.ts",
+    "test-coverage": "c8 --reporter=json --reporter=html npm test",
+    "test-staged": "npm test --bail",
```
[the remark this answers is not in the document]
justinmk: It's for husky or something like that. https://github.com/neovim/node-client/blob/4d6d4d3c4c79cd2fd09c039b3052223bddb86ef4/package.json#L44-L49

I'd probably prefer to remove lint-staged and similar stuff. This repo is simple and should stay simple, so not worth having that lint-staged dep imo.

packages/neovim/package.json
```diff
@@ -58,50 +58,27 @@
...
-    "@types/node": "^16.18.113",
-    "babel-jest": "^29.7.0",
-    "jest": "^29.7.0",
-    "jest-haste-map": "^29.7.0",
-    "jest-resolve": "^29.6.1",
+    "@types/mocha": "^10.0.9",
+    "@types/node": "16.9.x",
+    "c8": "^10.1.2",
+    "expect": "^29.7.0",
+    "jest-mock": "^29.7.0",
```
justinmk, line 65: `jest-mock` is fine as a temporary step (if it saves some noise in this PR), but maybe we want to switch to https://sinonjs.org/ ? Or maybe it doesn't matter. (But maybe it's confusing to have both jest and mocha :)

packages/neovim/package.json
```diff
@@ -58,50 +58,27 @@
...
     "@babel/preset-env": "^7.25.3",
     "@babel/preset-typescript": "^7.24.7",
-    "@types/jest": "^29.5.12",
-    "@types/node": "^16.18.113",
-    "babel-jest": "^29.7.0",
-    "jest": "^29.7.0",
-    "jest-haste-map": "^29.7.0",
-    "jest-resolve": "^29.6.1",
+    "@types/mocha": "^10.0.9",
+    "@types/node": "16.9.x",
```
justinmk, line 62: were the node types intentionally downgraded? they were `16.18.113` previously

.eslintrc.js
```diff
@@ -70,6 +71,7 @@ module.exports = {
 
     'import/extensions': 'off',
     'import/prefer-default-export': 'off',
+    "import/no-extraneous-dependencies": ["error", {"devDependencies": true, "optionalDependencies": false, "peerDependencies": false}],
```
justinmk, line 74: single quotes

packages/integration-tests/package.json
```diff
@@ -50,12 +50,10 @@
...
-    "jest": "^29.7.0",
-    "jest-haste-map": "^29.7.0",
-    "jest-resolve": "^29.6.1"
-  },
-  "jest": {
-    "testEnvironment": "node",
-    "testPathIgnorePatterns": [".d.ts", ".js"]
+    "@types/mocha": "^10.0.9",
+    "c8": "^10.1.2",
+    "expect": "^29.7.0",
```
justinmk, line 55: OK as a transitional step, but can probably be avoided in a later PR?

gjf7:
C8 requires node >= 18, wtf?

packages/neovim/src/testUtil.ts
```diff
@@ -19,8 +18,10 @@ export function findNvimOrFail() {
 
 const nvimPath = findNvimOrFail();
 
-let proc: cp.ChildProcessWithoutNullStreams;
-let nvim: NeovimClient;
+// eslint-disable-next-line import/no-mutable-exports
+export let proc: cp.ChildProcessWithoutNullStreams;
+// eslint-disable-next-line import/no-mutable-exports
+export let nvim: NeovimClient;
```
justinmk, line 24: can/should this be a function? it's a bit "magical" as a global var, and also makes debugging a bit more painful.

packages/neovim/src/host/NvimPlugin.test.ts
```diff
@@ -94,7 +95,8 @@ describe('NvimPlugin', () => {
   });
 
   it('should create functions from callable arrays', () => {
-    const fn = jest.fn(function () {
+    const fn = jestMock.fn(function () {
+      // @ts-expect-error intentionly
```
justinmk, line 99:
```suggestion
      // @ts-expect-error intentional
```

packages/neovim/src/host/NvimPlugin.test.ts
```diff
@@ -110,7 +112,8 @@ describe('NvimPlugin', () => {
       getFakeNvimClient()
     );
     const obj = {
-      func: jest.fn(function () {
+      func: jestMock.fn(function () {
+        // @ts-expect-error intentionly
```
justinmk, line 116:
```suggestion
        // @ts-expect-error intentional
```

justinmk (approved):
I was a bit worried about this but this looks like a win. You did it in a way that reduces risk and churn. Thank you!

justinmk:
> C8 requires node >= 18, wtf?

We could skip codecov for older node.

packages/neovim/src/testUtil.ts
gjf7: I'll handle this too in a follow up PR :)

packages/neovim/src/attach/attach.test.ts
```diff
@@ -155,7 +150,7 @@ describe('Nvim API', () => {
   });
 
   it('emits "disconnect" after quit', done => {
```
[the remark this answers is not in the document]
justinmk: we can skip it for now. even after the 'exit' workaround it still fails for me locally, sometimes.

gjf7:
looks like there is a Uncaught error in our test util.

gjf7:
> maybe drop `--parallel` for now? not sure if these tests are prepared for that.

Drop `--parallel` make `process.env.NODE_ENV = 'test' not executed, don't know why.

packages/neovim/src/testSetup.ts
```diff
@@ -1,12 +1,13 @@
 // Global test setup. Runs before each test.
+import { startNvim, stopNvim } from './testUtil';
 
-import * as testUtil from './testUtil';
-// import * as jest from '@jest/globals'
+process.env.NODE_ENV = 'test';
```
gjf7, line 4: Already done that, I mean this line is not working after drop --parallel.

gjf7:
Turns out we must set 'NODE_ENV=test' in .mocharc.js in non-parallel mode. Sets in testSetup is useless :)

gjf7:
Need to check why codecov is't triggered.

justinmk:
might need a `.c8rc.json` in packages/neovim/. IIRC the "lcov" reporter was important to get it to work with codecov.

example (untested, just a sample):

```
{
    "report-dir": "../../coverage/neovim",
    "reporter": ["lcov"],
    "all": true,
    "include": ["src/**"],
    "exclude": [
        "src/test/**"
    ]
}
```

package.json
```diff
@@ -30,7 +30,7 @@
     "dev": "npm run --stream --parallel dev --workspaces --if-present",
     "publish:neovim": "cd packages/neovim && cp ../../README.md . && cp ../../CHANGELOG.md . && npm publish",
     "test": "npm run test --workspaces --if-present",
-    "test-coverage": "npm run test --workspaces --if-present",
+    "test-coverage": "npm run test-coverage --workspaces --if-present",
```
gjf7, line 33: My bad :), codecov works now.

gjf7:
Follow up PR TODO:
1. eliminate the magical global `nvim`, `proc` in test util.
2. replace jest-mock with sinon, keep expect is fine, what do u think @justinmk ?

justinmk:
reading https://jestjs.io/docs/expect , mocha already provides this except the syntax is different?

.eslintrc.js
```diff
@@ -70,6 +71,7 @@ module.exports = {
 
     'import/extensions': 'off',
     'import/prefer-default-export': 'off',
+    'import/no-extraneous-dependencies': ['error', {'devDependencies': true, 'optionalDependencies': false, 'peerDependencies': false}],
```
justinmk, line 74: according to https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-extraneous-dependencies.md , `devDependencies:false` allows dev dependencies to be imported. We want that to be an error for non-test code.

To set this only for test code, put it in the `overrides.rules` (line 20 above)?

justinmk:
> replace jest-mock with sinon

On second thought, let's keep jest-mock for now. Mocking should be rare, and unless sinon improves something, using jest-mock avoids churn and doesn't really hurt.

.eslintrc.js
gjf7: Fixed

gjf7:
Ready for merge :)

PR #136, opened 2019-12-25
feat(transport): bump up msgpack to latest official

kwonoj:
This PR attempts to bump up msgpack from `msgpack-lite` to latest official (https://github.com/msgpack/msgpack-javascript).

msgpack-lite has not been updated for amount of time, meanwhile official bindings improved implementation, as well as some experimental perf improvement approach as well like wasm backend (https://github.com/msgpack/msgpack-javascript/blob/c969bc9045a2fc2e70cf3bddd7ccd2e71cff1a78/src/wasmFunctions.ts#L3, still experimental though)

to conform `@msgpack`'s api `transport`'s internal implementation has changed, but does not exposes any breaking changes so far.

PR #429, opened 2024-10-19
refactor: drop redundant buffered.ts

gjf7:


justinmk:
Thank you for noticing this! Looks like this was missed in https://github.com/neovim/node-client/pull/136

PR #435, opened 2024-11-01
build: infer prettier parser from file extension

mikavilpas:
Issue
=====

The prettier parser was set to `typescript` in the `.prettierrc.json` file. Because prettier was used for multiple files types, it caused errors when formatting files that were not TypeScript.

Here is an example:

```sh
README.md
[error] README.md: SyntaxError: Invalid character. (1:1)
[error] > 1 | # Neovim node.js client
[error]     | ^
[error]   2 |
[error]   3 | | CI (node >= 14, Linux/macOS/Windows) | Coverage | npm |
[error]   4 | |----------------------------|----------|-----|
```

This seems to cause the build to fail as of the most recent commit https://github.com/neovim/node-client/commit/74880148636620f3f76e0a568aed10ce18b512a1

Solution
========

Automatically infer the parser based on the file extension.

References:
- https://prettier.io/docs/en/configuration.html#setting-the-parserdocsenoptionshtmlparser-option

To make the build pass, run `npx prettier --write .` to format all files. I think it's best that someone else does this, as it will change the formatting of many files (currently 43), and reviewing them is very difficult.

### DIFF
--- a/.babelrc.js
+++ b/.babelrc.js
@@ -1,8 +1,5 @@
 module.exports = {
-  presets: [
-    ['@babel/preset-env', { targets: { node: '8.15.1' } }],
-    '@babel/preset-typescript',
-  ],
+  presets: [['@babel/preset-env', { targets: { node: '8.15.1' } }], '@babel/preset-typescript'],
   plugins: ['@babel/plugin-proposal-class-properties'],
 
   env: {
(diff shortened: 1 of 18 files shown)


mikavilpas:
Just rebased to master - I kept the new `printWidth` prettier setting you added in #444 

I also ran `npx prettier --write .` in another commit to keep things simple.

PR #437, opened 2024-11-01
chore: remove unneeded prettier eslint plugin

mikavilpas:
Issue
=====

Prettier rules seem to be enforced both via prettier itself, and in addition by eslint via eslint-plugin-prettier.

Prettier docs state the following recommendation:

> When searching for both Prettier and your linter on the Internet you’ll probably find more related
> projects. These are generally not recommended[...]
>
> By running Prettier inside your linters, you didn’t have to set up any new infrastructure and you
> could re-use your editor integrations for the linters.
> But these days you can run `prettier --check .` and most editors have Prettier support.
>
> The downsides of those plugins are:
>
> - You end up with a lot of red squiggly lines in your editor, which gets annoying. Prettier is
>   supposed to make you forget about formatting – and not be in your face about it!
>
> https://prettier.io/docs/en/integrating-with-linters.html

Solution
========

Remove eslint-plugin-prettier from the project, and continue to use prettier as a standalone tool. This removes the red squiggly lines in the editor.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -4,7 +4,7 @@ module.exports = {
     'plugin:@typescript-eslint/recommended',
     'prettier',
   ],
-  plugins: ['unicorn', 'import', 'prettier', '@typescript-eslint'],
+  plugins: ['unicorn', 'import', '@typescript-eslint'],
   parser: '@typescript-eslint/parser',
   parserOptions: {
     project: './tsconfig.json',
@@ -66,9 +66,6 @@ module.exports = {
     // Causes issues with enums
     'no-shadow': 'off',
     'prefer-destructuring': 'off',  // Intentionally disabled trash.
-
-    // prettier things
-    'prettier/prettier': 'error',
 
     'import/extensions': 'off',
     'import/prefer-default-export': 'off',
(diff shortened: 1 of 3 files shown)


justinmk:
Minimizing dependencies is definitely appreciated.

However in this case I don't understand what problem is being solved. In this project, the prettier rules *are* also enforced "lint" rules. That's intentional. I don't really give a flip about red squiggles, and I don't understand why that is undesirable anyways--because again, prettier rules *are lint rules*.

mikavilpas:
Oh, I did not realize that was the case. In my experience, usually prettier is run with an editor plugin, and on the command line in CI, while being run as part of eslint is an older way and nowadays not recommended.

If running as part of eslint is a good fit for this project, I think with the same reasoning https://github.com/neovim/node-client/pull/435 can also be closed. It's about some prettier errors that appear when you try to run pretttier on the command line.

justinmk:
> while being run as part of eslint is an older way and nowadays not recommended.

Ok, reviewing https://prettier.io/docs/en/integrating-with-linters.html , I guess I missed that.

But this PR doesn't add `prettier --check`, which means we are back to manually checking that prettier rules are enforced. 

I'd be in favor of removing the plugin as you did here, if we have:

- one `npm run lint` command that runs `prettier --check` + eslint
- one `npm run fixlint` command that fixes prettier issues + eslint issues

PR #441, opened 2024-11-04
style: fix eslint errors

mikavilpas:
Fixed by running
`eslint --format=unix --fix packages/*/src/**/*.ts packages/*/src/*.ts`

This is an alternative solution to #437 - it tries to fix the formatting by using eslint in order to make the build pass


justinmk:
thanks, I didn't see this until after https://github.com/neovim/node-client/pull/444 , which also updates `.prettierconfig`.

see also https://github.com/neovim/node-client/issues/445

Issue #481, opened 2025-08-24
Public method to close a client?


everett1992:
I'm looking for a method to disconnect and destroy a client when it is no longer needed. A client created from a socket will prevent the node process from exiting.

`nvim.transport.writer.end()` appears to work (at-least with a socket) but `transport` is private. I can also pass a socket as reader/writer and close that, but then I need to keep a references to the socket and client.

 Should there be `detatch` method on NeovimClient, or [Symbol.asyncDispose](https://v8.dev/features/explicit-resource-management) for `await using client = attach({ socket })`?

PR #482, opened 2025-08-28
feat: add public close methods

everett1992:
Add method to close transport and writer.

Before there was no public way to close the transport/reader/writer created by attach for a socket, and the connection would hold the node process open.

For example, this script would not exit (without an explicit process.exit(0))
```
attach({ socket: pathToNvimSocket })
```

This change adds .close to transport and NeovimClient which ends the writer. The other side of the connection should then close the reader which triggers detach and the cleanup code.

I added asyncDispose methods for folks using modern javascript runtimes that support explicit resource management:

```
await using client = attach({ socket: pathToNvimSocket })
```

Other wise you can call close
```
try {
  const client = attach({ socket: pathToNvimSocket })
} finally {
  await client.close()
}
```

fix https://github.com/neovim/node-client/issues/481

### DIFF
--- a/packages/neovim/src/api/client.ts
+++ b/packages/neovim/src/api/client.ts
@@ -6,6 +6,7 @@ import { Transport } from '../utils/transport';
 import { VimValue } from '../types/VimValue';
 import { Neovim } from './Neovim';
 import { Buffer } from './Buffer';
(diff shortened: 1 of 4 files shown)


justinmk:
This is a nice change, thanks!

